PR #2, opened 2020-06-24
bisection: Add an initial script for future bisection exercises.

tcwanggoogle:
The idea of the script is to count how many lines of comments are
there in the repo. So over time, when the repo grows, one comment
will make it over the set limit, to trigger the alert. This is
when you can start to use `git bisect` to find out the culprit.
In this way, you won't actually know which commit triggers the
exception by just looking at the commit contents (hopefully!)

Also, just in case you want to start earlier, or the number of
commits never go above the default value (1000), the script also
takes an argument to change the limit, so we can start the
exercise anytime.

As long as python is installed on your machine, the script should
be able to execute. Will seek other options if that's not working

Added a README as instructions.

### DIFF
--- a/bisection/application.py
+++ b/bisection/application.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python2
+
+"""The fake-core binary of your application
+
+Assume this is the core binary of your application. Assume your goal
+is to keep this binary running without any problems. The purpose is
+to train you using `git bisection` to find out which commit causes
+this binary stopped to work.
+
+The algorithm of this script is to find out how many lines of
+comments are there in the repo. If it exceeds the number we pass
+to this program, it'll raise an exception.
+"""
+
+from __future__ import print_function
+
+import argparse
+import os
+import sys
+
+COMMENT_STYLES = {
+    '.java': ['//', ('/*', '*/')],
+    '.js': ['//'],
+    '.html': [('<!--', '-->')],
+}
+
+def count_comments_from_contents(contents, style):
+    lines = 0
+    if isinstance(style, str):
+        # Consider single line comments
+        for line in contents.split('\n'):
+            if line.strip().startswith(style):
+                lines += 1
+        return lines
+
+    if isinstance(style, tuple):
+        # Consider multi-line comments
+        begin, end = style
+        in_comment_section = False
+        for line in contents.split('\n'):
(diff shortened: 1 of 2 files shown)


bisection/application.py
```diff
@@ -0,0 +1,92 @@
...
+to this program, it'll raise an exception.
+"""
+
+from __future__ import print_function
+
+import argparse
+import os
+import sys
+
+COMMENT_STYLES = {
```
TcgVanguardTroll, line 21: Should there be a inclusion for XML as is that covered in the same context as HTML ?

bisection/application.py
```diff
@@ -0,0 +1,92 @@
...
+    '.java': ['//', ('/*', '*/')],
```
kmh287, line 22: ('/**', '*/') should also capture javadocs

bisection/application.py
```diff
@@ -0,0 +1,92 @@
...
+    '.js': ['//'],
```
kmh287, line 23: /* */ and /** */ are also valid for js

bisection/application.py
tcwanggoogle: Just to be clear: only ('/**', '*/') is valid in Java or ('/*', '*/') is also valid?

kmh287: /** */ is used specifically for documentation e.g. class , method, or field docs.

/* */ just denotes a multi-line comment that can appear within the body of a method.

bisection/application.py
kmh287: `/**` `*/` and `/*` `*/`.

bisection/application.py
tcwanggoogle: Thanks for the clarification. If both are valid, then the current tuple ('/\*', '\*/') should be enough, as I only check the starting of the line. (If add two, javadocs will be counted twice)

bisection/application.py
tcwanggoogle: Same as above

bisection/application.py
tcwanggoogle: Thanks. Added the same comment style as html

tcwanggoogle:
Kevin, can you link the GCP project with the repo again (since the name changed)? Thanks!

PR #3, opened 2020-06-26
Add form with meeting time and duration input fields and Ad-lib content to UI

dhruvi618:
Next steps: Add more input fields to UI and retrieve form inputs on submit

### DIFF
--- a/frontend/src/menu-bar.js
+++ b/frontend/src/menu-bar.js
@@ -20,6 +20,7 @@ const useStyles = makeStyles(theme => ({
   }
 }));
 
+// Create menu bar component with title and help icon (with About and FAQ links)
 export default function MenuAppBar() {
   const classes = useStyles();
   const [anchorEl, setAnchorEl] = React.useState(null);
--- a/frontend/src/menu-bar.js
+++ b/frontend/src/menu-bar.js
@@ -1,0 +1,78 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
+import AppBar from "@material-ui/core/AppBar";
+import Toolbar from "@material-ui/core/Toolbar";
+import Typography from "@material-ui/core/Typography";
+import IconButton from "@material-ui/core/IconButton";
+import HelpIcon from "@material-ui/icons/Help";
+import MenuItem from "@material-ui/core/MenuItem";
+import Menu from "@material-ui/core/Menu";
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    flexGrow: 1
+  },
+  menuButton: {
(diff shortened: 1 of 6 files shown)


frontend/src/menu-bar.js
```diff
@@ -0,0 +1,79 @@
...
+  title: {
+    flexGrow: 1
+  }
+}));
+
+// Create menu bar component with title and help icon (with About and FAQ links)
+export default function MenuAppBar() {
+  const classes = useStyles();
+  const [anchorElement, setAnchorElement] = React.useState(null);
+  const open = Boolean(anchorElement);
```
kmh287, line 27: Probably a react question but why does this need to be wrapped in a Boolean? Could it be used directly, or is this used for a truthy --> boolean conversion?

frontend/yarn.lock
```diff
@@ -1069,7 +1069,7 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.5", "@babel/runtime@^7.5.1", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.4":
+"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.3.1", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.4", "@babel/runtime@^7.4.5", "@babel/runtime@^7.5.1", "@babel/runtime@^7.5.5", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.3", "@babel/runtime@^7.8.4", "@babel/runtime@^7.8.7":
```
tcwanggoogle, line 1072: Question: What causes this file to change? (just curious if we'll see this file change on every PR, or not)

frontend/src/menu-bar.js
dhruvi618: I realized I misinterpreted our original UI design for the menubar (was not supposed to have two options, just a click that redirects it to a FAQ page) so I got rid of this code, but this line could be use directly without the Boolean wrapper.

dhruvi618:
UI Design: https://screenshot.googleplex.com/Ra24iNGW7G1

frontend/yarn.lock
dhruvi618: This file changes when dependencies/libraries are added or updated. For example, in this PR, I added a material-ui icons import (to use the help icon) so it caused updates in the file. I don't think it should happen in every PR.

PR #5, opened 2020-06-26
Add matching algorithm and tests

jzwang8:
I'm taking suggestions for more tests for FindMatchingQuery! Also, the Match, Participant, and Time are in other PRs so you may have to reference those to understand this chunk.

### DIFF
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -1,0 +1,82 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import com.google.gson.Gson;
+import com.google.sps.data.Match;
+import com.google.sps.data.Participant;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class FindMatchQuery {
+
+  private Date date;
+  
+  /** Constructor */
+  public FindMatchQuery() {
+    date = new Date();
+  }
+
+  /** Constructor with manually set date*/
+  public FindMatchQuery(Date date) {
+    this.date = date;
+  }
+
+  /** Find match of new participant or add to list of participants, return whether or not match was found 
+      right after being added */
+  public Match findMatchQuery(List<Participant> participants) {
+    
+    // Compare new participant preferences with others in list to find match
+    int numParticipants = participants.size();
+    Participant newParticipant = participants.get(numParticipants - 1);
+    for (int i = 0; i < numParticipants - 1; i++) {
+      Participant currParticipant = participants.get(i);
+      
+      // Check if participants are looking for similar meeting duration
+      int newDuration = newParticipant.getDuration();
+      int currDuration = currParticipant.getDuration();
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= 15;
+      int duration = Math.min(newDuration, currDuration);
+      
+      if (compatibleDuration) {
+        // Check if participants are both free for that duration + extra
+        long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
+        long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
+        boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + 15).toMillis() <=
+                                    Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+        
+        if (compatibleTime) {
+          // TODO: change match ID (currently -1 for easy error checking)
+          return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
+        }
+        else {
+          continue;
+        }
+      }
+      else {
+        continue;
+      }
+    }
+    // No inital match found
+    return null;
+  }
+}
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,12 +21,12 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
-import java.util.List;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
+/** Class used to find a match in a list of Participants with the most recently added Participant */
 public final class FindMatchQuery {
 
   private Date date;
@@ -36,44 +36,40 @@ public final class FindMatchQuery {
     date = new Date();
   }
 
-  /** Constructor with manually set date*/
+  /** Constructor with manually set date */
   public FindMatchQuery(Date date) {
     this.date = date;
   }
 
   /** Find match of new participant or add to list of participants, return whether or not match was found 
-      right after being added */
-  public Match findMatchQuery(List<Participant> participants) {
+    * right after being added
+    */
+  public Match findMatch(List<Participant> participants, Participant newParticipant) {
+    int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
+    int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
     
     // Compare new participant preferences with others in list to find match
-    int numParticipants = participants.size();
-    Participant newParticipant = participants.get(numParticipants - 1);
-    for (int i = 0; i < numParticipants - 1; i++) {
-      Participant currParticipant = participants.get(i);
-      
+    for (Participant currParticipant : participants) {
+
       // Check if participants are looking for similar meeting duration
       int newDuration = newParticipant.getDuration();
       int currDuration = currParticipant.getDuration();
-      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= 15;
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
       int duration = Math.min(newDuration, currDuration);
       
-      if (compatibleDuration) {
-        // Check if participants are both free for that duration + extra
-        long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
-        long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
-        boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + 15).toMillis() <=
-                                    Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
-        
-        if (compatibleTime) {
-          // TODO: change match ID (currently -1 for easy error checking)
-          return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
-        }
-        else {
-          continue;
-        }
-      }
-      else {
+      if (!compatibleDuration) {
         continue;
+      }
+      
+      // Check if participants are both free for that duration + extra
+      long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
+      long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
+      boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + PADDING_TIME).toMillis() <=
+                                  Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+      
+      if (!compatibleTime) {
+        // TODO: change match ID (currently -1 for easy error checking)
+        return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
       }
     }
     // No inital match found
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -17,7 +17,9 @@ package com.google.sps;
 import com.google.gson.Gson;
 import com.google.sps.data.Match;
 import com.google.sps.data.Participant;
+import java.time.Clock;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -29,18 +31,18 @@ import java.util.stream.Collectors;
 /** Class used to find a match in a list of Participants with the most recently added Participant */
 public final class FindMatchQuery {
 
-  private Date date;
+  private Clock clock;
   private static int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
   private static int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
   
   /** Constructor */
   public FindMatchQuery() {
-    date = new Date();
+    clock = new Clock();
   }
 
   /** Constructor with manually set date */
-  public FindMatchQuery(Date date) {
-    this.date = date;
+  public FindMatchQuery(Clock clock) {
+    this.clock = clock;
   }
 
   /** Find match of new participant or add to list of participants, return whether or not match was found 
@@ -64,8 +66,9 @@ public final class FindMatchQuery {
       // Check if participants are both free for that duration + extra
       long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
       long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
-      boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + PADDING_TIME).toMillis() <=
-                                  Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+      boolean compatibleTime = Instant.now(clock).plusMillis(Duration.ofMinutes(duration + PADDING_TIME).toMillis())
+          .isBefore(Instant.ofEpochMillis(Math.min(newTimeAvailableUntil, currTimeAvailableUntil)));
+
       if (compatibleTime) {
         // TODO: change match ID (currently -1 for easy error checking)
         return new Match(/* id= */ -1L, /* firstParticipant= */ newParticipant, /* secondParticipant= */ currParticipant, /* duration= */ duration, /* timestamp= */ date.getTime());
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -30,6 +30,8 @@ import java.util.stream.Collectors;
 public final class FindMatchQuery {
 
   private Date date;
+  private static int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
+  private static int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
   
   /** Constructor */
   public FindMatchQuery() {
@@ -45,8 +47,6 @@ public final class FindMatchQuery {
     * right after being added
     */
   public Match findMatch(List<Participant> participants, Participant newParticipant) {
-    int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
-    int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
     
     // Compare new participant preferences with others in list to find match
     for (Participant currParticipant : participants) {
@@ -64,12 +64,11 @@ public final class FindMatchQuery {
       // Check if participants are both free for that duration + extra
       long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
       long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
-      boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + PADDING_TIME).toMillis() <=
-                                  Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+      boolean compatibleTime = Instant.now().plusMillis(Duration.ofMinutes(duration + PADDING_TIME).toMillis()).isBefore(Math.min(newTimeAvailableUntil, currTimeAvailableUntil));
       
       if (compatibleTime) {
         // TODO: change match ID (currently -1 for easy error checking)
-        return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
+        return new Match(/* id= */ -1L, /* firstParticipant= */ newParticipant, /* secondParticipant= */ currParticipant, /* duration= */ duration, /* timestamp= */ date.getTime());
       }
     }
     // No inital match found
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -18,6 +18,7 @@ import com.google.gson.Gson;
 import com.google.sps.data.Match;
 import com.google.sps.data.Participant;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -64,7 +65,7 @@ public final class FindMatchQuery {
       // Check if participants are both free for that duration + extra
       long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
       long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
-      boolean compatibleTime = Instant.now().plusMillis(Duration.ofMinutes(duration + PADDING_TIME).toMillis()).isBefore(Math.min(newTimeAvailableUntil, currTimeAvailableUntil));
+      boolean compatibleTime = Instant.now().plusMillis(Duration.ofMinutes(duration + PADDING_TIME).toMillis()).isBefore(Instant.ofEpochMilli(Math.min(newTimeAvailableUntil, currTimeAvailableUntil)));
       
       if (compatibleTime) {
         // TODO: change match ID (currently -1 for easy error checking)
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.sps.data.Match;
+import com.google.sps.data.Participant;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
```
dhruvi618, line 27: nit: two imports for java.util.List

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+        if (compatibleTime) {
+          // TODO: change match ID (currently -1 for easy error checking)
+          return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
+        }
+        else {
+          continue;
+        }
+      }
+      else {
+        continue;
```
dhruvi618, line 76: Since the continue statements are at the end of the loop and there aren't other lines of execution afterwards, I think it may be possible to remove them.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+  public void threeParticipants23() {
+    /* Three participants, 1st and 2nd aren't compatible, but 2nd and 3rd are */
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0250PM, TIMEZONE_DEFAULT, DURATION_45_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match actual = query.findMatchQuery(Arrays.asList(participantA, participantB, participantC));
+    
+    assertThat(actual.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(actual.getSecondParticipant().getLdap().equals(PERSON_A));
```
dhruvi618, line 132: Should this be PERSON_B since participants 2 and 3 are compatible?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+    assertThat(actual.getDuration() == DURATION_60_MINUTES);
+  }
+
+  @Test
+  public void fourParticipantsTwoMatches() {
+    /* Three participants, 1st and 2nd aren't compatible, but 2nd and 3rd are */
```
dhruvi618, line 138: tiny nit: "Four participants, 1st and 2nd are compatible"

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match actual = query.findMatchQuery(Arrays.asList(participantA, participantB, participantC));
+    
+    assertThat(actual.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(actual.getSecondParticipant().getLdap().equals(PERSON_A));
+    assertThat(actual.getDuration() == DURATION_60_MINUTES);
+  }
+
+  @Test
```
dhruvi618, line 136: What do you think about adding a test where there are three participants (A,B,C) where both A-C and B-C match, but the algorithm only returns back one of these matches?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+public final class FindMatchQuery {
+
+  private Date date;
+  
+  /** Constructor */
+  public FindMatchQuery() {
+    date = new Date();
+  }
+
+  /** Constructor with manually set date*/
```
dhruvi618, line 39: tiny nit: space after date

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class FindMatchQuery {
+
+  private Date date;
+  
+  /** Constructor */
```
dhruvi618, line 34: I think the `/**` makes this a javadoc instead of a comment

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
+// Copyright 2019 Google LLC
```
kmh287, line 1: 2020

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class FindMatchQuery {
+
+  private Date date;
```
kmh287, line 32: Can you add javadocs to explain how this will be used?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+    date = new Date();
+  }
+
+  /** Constructor with manually set date*/
+  public FindMatchQuery(Date date) {
+    this.date = date;
+  }
+
+  /** Find match of new participant or add to list of participants, return whether or not match was found 
+      right after being added */
```
kmh287, line 45: nit: multiline javadocs put the stars on their own lines so

```
/**
  * 
  */
```

instead of 

```
/** 
   ........ */
```

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+  public Match findMatchQuery(List<Participant> participants) {
+    
+    // Compare new participant preferences with others in list to find match
+    int numParticipants = participants.size();
+    Participant newParticipant = participants.get(numParticipants - 1);
+    for (int i = 0; i < numParticipants - 1; i++) {
```
kmh287, line 51: can this be a for-each loop over `participants`? Doing so allows the loop to use `participants`'s iterator instead of calls to get which may not run in constant time.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+    this.date = date;
+  }
+
+  /** Find match of new participant or add to list of participants, return whether or not match was found 
+      right after being added */
+  public Match findMatchQuery(List<Participant> participants) {
+    
+    // Compare new participant preferences with others in list to find match
+    int numParticipants = participants.size();
+    Participant newParticipant = participants.get(numParticipants - 1);
```
kmh287, line 50: Could `newParticipant` be separated out into a second parameter alongside `participants`? Doing so can remove this assumption about the structure of `participants`

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+        boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + 15).toMillis() <=
+                                    Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+        
+        if (compatibleTime) {
+          // TODO: change match ID (currently -1 for easy error checking)
+          return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
+        }
+        else {
+          continue;
+        }
```
kmh287, line 73: no need for `else continue` since the if returns and there's no more code in the loop after this

backend/src/main/java/com/google/sps/FindMatchQuery.java
kmh287: +1. `continue` could be useful here to help unindent the main body of the loop above. Instead of 

```
for (....) {
  if (condition) {
    // Large clock of code
  }
}
```

You can unindent the inner block with 

```
for (...) {
  if (!condition) {
    continue;
  }
  // Large block of code
}

File: backend/src/test/java/com/google/sps/FindMatchQueryTest.java [hunk 14]
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Calendar;
+import java.util.Objects;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import static org.mockito.Mockito.*;
+
+/** */
```
kmh287, line 32: It's okay to omit docs for test classes so you can delete the line entirely.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+  private static final int DURATION_60_MINUTES = 60;
+
+  private FindMatchQuery query;
+  private Time time;
+
+  @Before
+  public void setUp() {
+    // Set "current" date to  1/1/2020 2:00pm GMT
+    Calendar c = Calendar.getInstance();
+    c.set(2020, 0, 1, 14, 0, 0);
```
kmh287, line 65: /* param= */ for each of the literals

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+    time = new Time(date);
+  }
+
+  @Test
+  public void compatibleTimeAndDuration() {
+    /* Two participants who are compatible in available time AND duration */
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_15_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match actual = query.findMatchQuery(Arrays.asList(participantA, participantB));
```
kmh287, line 79: here and below:

internally we have a convention to use ImmutableList from Google collections when passing lists in tests. Not only is the API slightly shorter, its use also helps catch bugs where methods unexpectedly mutate lists.

kmh287, line 79: Since there's no `expected`, I'd recommend renaming `actual` to `match`

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+    assertThat(actual.getFirstParticipant().getLdap().equals(PERSON_B));
+    assertThat(actual.getSecondParticipant().getLdap().equals(PERSON_A));
+    assertThat(actual.getDuration() == DURATION_15_MINUTES);
+  }
+
+  @Test
+  public void compatibleTime() {
+    /* Two participants who are compatible in available time but NOT duration */
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
```
kmh287, line 90: why can't they meet for 30 minutes?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,151 @@
...
+    
+    Match actual = query.findMatchQuery(Arrays.asList(participantA, participantB));
+
+    assertThat(Objects.isNull(actual));
```
kmh287, line 94: assertThat(actual).isNull()

As written this assertion is only partially complete and so the test will pass even if `actual` isn't null. To assert on booleans you would write

`assertThat(myBoolean).isTrue()` or `assertThat(myBoolean).isFalse()`

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+    // Compare new participant preferences with others in list to find match
+    int numParticipants = participants.size();
+    Participant newParticipant = participants.get(numParticipants - 1);
+    for (int i = 0; i < numParticipants - 1; i++) {
+      Participant currParticipant = participants.get(i);
+      
+      // Check if participants are looking for similar meeting duration
+      int newDuration = newParticipant.getDuration();
+      int currDuration = currParticipant.getDuration();
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= 15;
```
kmh287, line 57: Please extract `15` to a private static final constant with a name that explains its use like "MINIMUM_MEETING_TIME". Doing so helps future readers of your code understand the significance of the value and also makes it easy to change if it's referenced multiple times in the code (as it is here)

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /** Constructor with manually set date*/
+  public FindMatchQuery(Date date) {
+    this.date = date;
+  }
+
+  /** Find match of new participant or add to list of participants, return whether or not match was found 
+      right after being added */
+  public Match findMatchQuery(List<Participant> participants) {
```
tcwanggoogle, line 46: Super tiny nit: This function name is almost the same as the constructor (with only the leading character different). I worry it might be error prone. Should we rename it?

backend/src/main/java/com/google/sps/FindMatchQuery.java
tcwanggoogle: +1 to this suggestion

backend/src/main/java/com/google/sps/FindMatchQuery.java
tcwanggoogle: +1

backend/src/main/java/com/google/sps/FindMatchQuery.java
tcwanggoogle: +1

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: The 15s are actually referring to different things, but I made constants for both for clarity.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Got it.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Their desired meeting durations are not within 15 minutes of each other. Just a rule we made up to determine if they're a match or not. 30&60, 15&45 are not compatible durations.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Great suggestion, (I think) I made this change? I just did query.findMatchQuery(ImmutableList.copyOf(Arrays.asList(participantA, participantB)));

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Made it so that this method takes in two parameters: a list of current participants, and the new participant. Now the participants loop doesn't loop through just the first numParticipants - 1

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Thanks for the suggestion, added! threeParticipantsTwoMatches().

backend/src/main/java/com/google/sps/FindMatchQuery.java
kmh287: not done

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,78 @@
...
+    for (Participant currParticipant : participants) {
+
+      // Check if participants are looking for similar meeting duration
+      int newDuration = newParticipant.getDuration();
+      int currDuration = currParticipant.getDuration();
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
+      int duration = Math.min(newDuration, currDuration);
+      
+      if (!compatibleDuration) {
+        continue;
```
kmh287, line 61: this is very clear and easy to follow 👍

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,78 @@
...
+  public FindMatchQuery(Date date) {
+    this.date = date;
+  }
+
+  /** Find match of new participant or add to list of participants, return whether or not match was found 
+    * right after being added
+    */
+  public Match findMatch(List<Participant> participants, Participant newParticipant) {
+    int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
+    int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
```
kmh287, line 49: These could be private static final constants so they aren't assigned each time the method runs.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,78 @@
...
+      int duration = Math.min(newDuration, currDuration);
+      
+      if (!compatibleDuration) {
+        continue;
+      }
+      
+      // Check if participants are both free for that duration + extra
+      long newTimeAvailableUntil = newParticipant.getTimeAvailableUntil();
+      long currTimeAvailableUntil = currParticipant.getTimeAvailableUntil();
+      boolean compatibleTime = date.getTime() + Duration.ofMinutes(duration + PADDING_TIME).toMillis() <=
```
kmh287, line 67: boolean compatibleTime = Instant.now().plusMillis(Duration ....).isBefore(......)

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,78 @@
...
+                                  Math.min(newTimeAvailableUntil, currTimeAvailableUntil);
+      
+      if (compatibleTime) {
+        // TODO: change match ID (currently -1 for easy error checking)
+        return new Match(-1L, newParticipant, currParticipant, duration, date.getTime());
```
kmh287, line 72: /* param= */

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
kmh287: Got it, thanks

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Test
+  public void compatibleTime() {
+    // Two participants who are compatible in available time but NOT duration
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match match = query.findMatch(Arrays.asList(participantA), participantB);
+
+    assertThat(Objects.isNull(match)).isTrue();
```
kmh287, line 94: assertThat(match).isNull() here and below

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,149 @@
...
+    // Three participants, 1st & 2nd, 1st & 3rd are compatible but only return 1st & 3rd
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_15_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0800PM, TIMEZONE_DEFAULT, DURATION_45_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match match = query.findMatch(Arrays.asList(participantA, participantB), participantC);
+    
+    assertThat(match.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(match.getSecondParticipant().getLdap().equals(PERSON_A));
+    assertThat(match.getDuration() == DURATION_60_MINUTES);
```
kmh287, line 147: assertThat(match.getDuration()).isEqualTo(DURATION_60_MINUTES) otherwise this is an incomplete assertion

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,149 @@
...
+    // Three participants, 1st and 2nd aren't compatible, but 2nd and 3rd are
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0250PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match match = query.findMatch(Arrays.asList(participantA, participantB), participantC);
+    
+    assertThat(match.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(match.getSecondParticipant().getLdap().equals(PERSON_B));
+    assertThat(match.getDuration() == DURATION_60_MINUTES);
```
kmh287, line 133: isEqualTo

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,149 @@
...
+    // Three participants, 1st and 2nd aren't compatible, but 1st and 3rd are
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0250PM, TIMEZONE_DEFAULT, DURATION_45_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_60_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match match = query.findMatch(Arrays.asList(participantA, participantB), participantC);
+    
+    assertThat(match.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(match.getSecondParticipant().getLdap().equals(PERSON_A));
+    assertThat(match.getDuration() == DURATION_60_MINUTES);
```
kmh287, line 119: isEqualTo

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: I don't think I can use Instant.now() here because for testing this class, I have to create a new reference Date. Any suggestions on how to refactor all of the Date/time code to be more consistent and easier to understand/use? All of the date objects I've found are sort of confusing.

kmh287: Using Instant.now() would eliminate the need to use Date in this class and in the test. The test has control over the meeting duration and the `timeAvailableUntil` values which should give enough flexibility to test this statement.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,149 @@
...
+    // Three participants, 1st & 2nd, 1st & 3rd are compatible but only return 1st & 3rd
+    Participant participantA = new Participant(ID_DEFAULT, PERSON_A, TIME_0400PM, TIMEZONE_DEFAULT, DURATION_30_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantB = new Participant(ID_DEFAULT, PERSON_B, TIME_0600PM, TIMEZONE_DEFAULT, DURATION_15_MINUTES, TIMESTAMP_DEFAULT);
+    Participant participantC = new Participant(ID_DEFAULT, PERSON_C, TIME_0800PM, TIMEZONE_DEFAULT, DURATION_45_MINUTES, TIMESTAMP_DEFAULT);
+    
+    Match match = query.findMatch(Arrays.asList(participantA, participantB), participantC);
+    
+    assertThat(match.getFirstParticipant().getLdap().equals(PERSON_C));
+    assertThat(match.getSecondParticipant().getLdap().equals(PERSON_A));
+    assertThat(match.getDuration()).isEqualTo(DURATION_30_MINUTES);
```
kmh287, line 147: The first two assertions are not complete and this will always pass. When using the Truth library you call `assertThat()` on the thing you want to assert about, and then call a second method to specify your assertion e.g. isEqualTo, isNotNull, isLessThan etc. Without that second part, the test won't assert anything. 

Please take a pass over all the assertions in this file and make sure they have the format 

`assertThat(...).isEqualTo(....)`

jzwang8: Fixed all.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: I changed FindMeetingQuery to depend on Clock and will change other files later.

PR #6, opened 2020-06-26
Add servlets to use query algorithm

jzwang8:
I haven't tested these yet.


backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,125 @@
...
+    matchEntity.setProperty("secondParticipant", match.getSecondParticipant());
+    matchEntity.setProperty("duration", match.getDuration());
+    matchEntity.setProperty("timestamp", match.getTimestamp());
+
+    // Insert entity into datastore
+    datastore.put(matchEntity);
+  }
+
+  /** Delete matched participants from datastore */
+  private void deleteParticipantsFromList(Match match, DatastoreService datastore) {
```
dhruvi618, line 119: Maybe rename to `deleteParticipantsFromDatastore`?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -0,0 +1,42 @@
+package com.google.sps.data;
+
+/** A user who want to be matched. */
+public final class Participant {
+
+  private final long id;
+  private final String ldap; // Google username
```
tcwanggoogle, line 7: Like mentioned in PR #9, I'd suggest to rename `ldap` to `username`

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,125 @@
...
+    participantEntity.setProperty("duration", duration);
+    participantEntity.setProperty("timestamp", timestamp);
+
+    // Insert entity into datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(participantEntity);
+
+    // Find match if possible or add to datastore.
+    FindMatchQuery findMatchQuery = new FindMatchQuery();
+    Match match = findMatchQuery.findMatchQuery(getParticipants());
```
tcwanggoogle, line 71: nit: I'd avoid naming the object the same as the class method, so this function call is somewhat weird by the looking. (Since it's only needed in getting `match`, maybe just call it `query`?)

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,82 @@
...
+        long id = (long) entity.getKey().getId();
+        Participant firstParticipant = (Participant) entity.getProperty("firstParticipant");
+        Participant secondParticipant = (Participant) entity.getProperty("secondParticipant");
+        int duration = (int) entity.getProperty("duration"); 
+        long timestamp = (long) entity.getProperty("timestamp");
+        Match match = new Match(id, firstParticipant, secondParticipant, duration, timestamp);
+        matches.add(match);
+    }
+
+    Match matchDetails = new Match(0, null, null, 0, System.currentTimeMillis()); // default if no match
```
tcwanggoogle, line 67: nit: Maybe create a new constructor in Match only taking current time to construct Match (and initialize all other fields to zero)?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,82 @@
...
+    // Brute force search for match
+    for (Match match : matches) {
+      if (ldap.equals(match.getFirstParticipant().getLdap()) || 
+            ldap.equals(match.getSecondParticipant().getLdap())) {
```
tcwanggoogle, line 71: Indent? (Or is it the right indentation in Java?)

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,125 @@
...
+    if (match != null) {
+      addMatchToDatastore(match, datastore);
+      deleteParticipantsFromList(match, datastore);
+    }
```
dhruvi618, line 75: What do you think about moving this part (and the corresponding methods it calls) to the `SearchMatchServlet` to separate out the functionalities of the post and finding/searching for a match? I feel like this can help in modularization and may also be able to eliminate the "brute force search for a match" on line 68 of the `SearchMatchServlet` since you will be doing this part there instead. On the front-end, I will make sure to both post to the `AddParticipantServlet` and also fetch information about a match from the `SearchMatchServlet` every time the form is submitted.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,82 @@
...
+    String ldap = email.split("@")[0];
+
+    // Create and sort queries by time
+    Query query = new Query("Match").addSort("timestamp", SortDirection.DESCENDING);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    // Convert list of entities to list of matches
+    List<Match> matches = new ArrayList<Match>();
```
dhruvi618, line 56: What do you think about creating a separate Match object for the purposes of sending/returning data (to the front-end or for notification purposes) or maybe limiting the fields this one holds to reduce overhead since each functionality requires different data to be returned

jzwang8: I'm not sure I understand what you mean. Could you please clarify?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: The reason why I have finding a match in the AddParticipantServlet part is because when you add a new participant, our algorithm calls for immediately checking if that new participant has a match. If not, then they're added to the datastore. If they do, then they're never added to the datastore of participants and just added as a match to the match datastore that's then searched in SearchMatchServlet. I don't think it would make sense to have finding and searching together.

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: I'll make this change after all PRs are approved since ldap is used across all files and would be confusing to change in only this PR.

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -0,0 +1,42 @@
...
+  private final long timeAvailableUntil; // Time user is available until
```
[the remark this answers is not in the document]
jzwang8: The timestamp is when the participant is "created" in AddParticipantServlet. So that's the beginning of their availability range. If you're talking about the beginning of the meeting time, that will be calculated when we create the calendar event based on the duration and the time that the match is found.

backend/src/main/java/com/google/sps/data/Participant.java
tcwanggoogle: sg

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Request parameter values
+    // TODO: Error check input values (before allowing to submit)
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    String ldap = email.split("@")[0];
+    long timeAvailableUntil = Long.parseLong(request.getParameter("timeAvailableUntil"));
+    String timezone = request.getParameter("timezone");
+    int duration = Integer.parseInt(request.getParameter("duration"));
```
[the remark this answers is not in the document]
jzwang8: Added error checking when parsing.

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -0,0 +1,50 @@
...
+    this.timestamp = timestamp;
+  }
+  
+  /** Constructor for empty match except timestamp*/
+  public Match(long timestamp) {
+    id = 0L;
+    firstParticipant = null;
+    secondParticipant = null;
+    duration = 0;
+    this.timestamp = timestamp;
```
tcwanggoogle, line 27: Optional: call full arguments version of Match() by passing all the zero variables.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,150 @@
...
+*/
+@WebServlet("/add-participant")
+public class AddParticipantServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Request parameter values
+    // TODO: Error check input values (before allowing to submit)
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
```
TcgVanguardTroll, line 50: What happens in the case this returns null ?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,150 @@
...
+    String ldap = email.split("@")[0];
```
TcgVanguardTroll, line 51: Is there a reason the 0 is in brackets as well as not within thee. split call, split takes two arguments the reg-ex and the amount you wish to split by. Since we know this is an email and there can only be 1 @ symbol I think you can do **_email.split("a",1)_**

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,81 @@
...
+*/
+@WebServlet("/search-match")
+public class SearchMatchServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get participant ldap
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    String ldap = email.split("@")[0];
```
TcgVanguardTroll, line 47: Same question as above

backend/src/main/java/com/google/sps/data/Participant.java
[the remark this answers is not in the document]
kmh287: I think it would be good to distinguish between when the entry was created and the beginning of the participant's availability range. In the future if you want to allow users to say "I'm not free now, but I would like to set something up two hours from now" then supporting a distinct start of the range now would make that easier.

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -0,0 +1,42 @@
...
+  private final String timezone; // User timezone
+  private final int duration; // How long user wants to chat
+  private final long timestamp; // Time of submitted form
```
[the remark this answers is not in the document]
kmh287: Each of these comments could be the javadocs for the field. So instead of 

```
  private final int duration; // How long user wants to chat
  private final long timestamp; // Time of submitted form
```

It could be
```
/** How long the user wants to chat. */
private final int duration;
/** Time of submitted form. */
private final long timestamp;
```

This is the standard way of documenting Java fields and allows tools to auto-generate docs pages for your classes. It also helps the IDE show hints when fields are highlighted with the cursor or referenced elsewhere.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
kmh287: This looks like +6 but it should be +4. Now that Dhruvi has submitted the pre-commit hook it should catch and auto-format cases like this.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,150 @@
...
+    long timeAvailableUntil = convertToLong(request.getParameter("timeAvailableUntil"));
+    String timezone = request.getParameter("timezone");
+    int duration = convertToInt(request.getParameter("duration"));
```
kmh287, line 54: Now that these could return -1 on invalid input, the servlet should stop processing the request here and return an error to the client if either timeAvailableUntil or duration is invalid.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,150 @@
...
+    long timestamp = System.currentTimeMillis();
+
+    Participant newParticipant = new Participant(-1L, ldap, timeAvailableUntil, timezone, duration, timestamp);
```
kmh287, line 57: /* param= */ on the literal. 

I think a one-line comment explaining why -1 is used might also be helpful

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
dhruvi618: Yes. For the notifications part, we need to provide it with the two usernames/ldaps of the people who are matched and for the frontend we need to provide it a match status (true/false on whether or not it is a match), so I was thinking that instead of returning the entire match object, we could instead return just the two usernames or a true/false so we are not sending back extra data that we are not planning to use. (I think we could make it even simpler by sending back the two matched usernames in either case as well).

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: The second argument is optional, so I just omitted it.

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -0,0 +1,46 @@
+package com.google.sps.data;
+
+/** A match of two users */
+public final class Match {
+
+  private final long id;
+  private final Participant firstParticipant; /** First participant */
+  private final Participant secondParticipant; /** Second participant */
+  private final int duration; /** Duration of meeting */
+  private final long timestamp; /** Time match found */
```
kmh287, line 10: javadoc comments should go above the thing they document.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Also, there's a [0] because I'm taking the first element of the array (ldap). Example: hello@gmail.com is split into ["hello", "gmail.com"], and I just want the "hello" part.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,81 @@
...
+        matchDetails = match;
+        break;
+      }
+    }
+    
+    // Send the JSON back as the response
+    Gson gson = new Gson();
+    response.setContentType("application/json");
+    response.getWriter().println(gson.toJson(matchDetails));
+  }
```
kmh287, line 80: I'm concerned about using an instance of the Match class to indicate no Match here. The second constructor in Match that takes a timestamp adds the possibility that any of the fields on Match are null / zero values. So any caller to match.getFirstParticipant() would have to check for null because it's possible and expected for some Matches to have null participant fields.

Could another object or format be used to communicate to the client that no match was found?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Added an error response to catch invalid input.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Added an error response to catch null email.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,151 @@
...
+
+    // Redirect back to the HTML page
+    response.sendRedirect("/index.html");
+  }
+
+  /** Return list of current participants from datastore*/
+  private List<Participant> getParticipants(DatastoreService datastore) {
+
+    // Create and sort participant queries by time
+    Query query = new Query("Participant").addSort("timestamp", SortDirection.DESCENDING);
```
kmh287, line 92: is there any way to narrow this query down to participants available within a desired time range? For a sufficiently large set of participants, loading the entire table into memory could be expensive, even prohibitively so.

backend/src/main/java/com/google/sps/data/Participant.java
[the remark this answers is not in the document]
jzwang8: I wrote this down as a todo after these PRs get merged. Because my PRs are all interdependent, it would be too complicated to refactor everything in this PR, but I will definitely do this in the next one.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,151 @@
...
+    matchEntity.setProperty("secondParticipant", match.getSecondParticipant());
+    matchEntity.setProperty("duration", match.getDuration());
+    matchEntity.setProperty("timestamp", match.getTimestamp());
+
+    // Insert entity into datastore
+    datastore.put(matchEntity);
+  }
+
+  /** Delete matched participants from datastore */
+  private void deleteSecondParticipantFromDatastore(Participant secondParticipant, DatastoreService datastore) {
```
kmh287, line 124: s/secondParticipant/participant

Where this method is used, secondParticipant is the variable used but this method doesn't know anything about first vs second

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,151 @@
...
+    Key secondParticipantEntityKey = KeyFactory.createKey("Participant", secondParticipant.getId());
+    datastore.delete(secondParticipantEntityKey);
+  }
+
+
+  /** Return positive long value, or -1L if invalid or negative */
+  private static long convertToLong(String s) {
```
kmh287, line 131: convertToPositiveLong since this method doesn't allow for negative values.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+    try {
+        long parsed = Long.parseLong(s);
+        return (parsed >= 0L) ? parsed : -1L;
+    } catch(NumberFormatException e) { 
+        return -1L;
+    }
+  }
+  /** Return positive integer value, or -1 if invalid or negative */
+  private static int convertToInt(String s) {
```
kmh287, line 143: same comment

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    Match matchDetails = new Match(System.currentTimeMillis()); // default if no match
+    // Brute force search for match
+    for (Match match : matches) {
+      if (ldap.equals(match.getFirstParticipant().getLdap()) || ldap.equals(match.getSecondParticipant().getLdap())) {
+        matchDetails = match;
+        break;
+      }
+    }
```
kmh287, line 74: There's an opportunity for the stream API here
```
Match matchDetails = matches.stream().filter(...).findFirst() 
if (matchDetails == null) {
   // handling for no match....
}
```

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
jzwang8: Got it, now it returns:
- matchStatus: false if it doesn't exists
- matchStatus: true, firstParticipantLdap: __, secondParticipantLdap:__, duration: __ if it does

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
jzwang8: I changed this part so it returns:
- matchStatus: false if it doesn't exists
- matchStatus: true, firstParticipantLdap: xx, secondParticipantLdap: xx, duration: xx if it does. 
Does this work better? Now they'd just check the value of matchStatus.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: The only way I can think of narrowing this query down is by moving determination of compatibility based on timeAvailableUntil and duration into this method instead of in FindMatchQuery. Should I still do that? It doesn't really make sense when the matching alg only involves time compatibility but seems to make more sense once I add in personal fields (role, product area, etc.)

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
jzwang8: I changed the response type so that it doesn't ever return a Match as a json object but instead varies depending on whether a match exists or not. Do you think using the stream API would still be helpful here?

jzwang8: I guess my question is what's the benefit of using the stream API besides making the for each loop one statement (but still multiple lines)

PR #8, opened 2020-06-26
Add Time class and tests

jzwang8:
Time class used in FindMatchQuery tests.

### DIFF
--- a/backend/src/main/java/com/google/sps/data/Time.java
+++ b/backend/src/main/java/com/google/sps/data/Time.java
@@ -3,28 +3,27 @@ package com.google.sps.data;
 import java.util.Date;
 import java.util.Calendar;
 
-/** Class to convert Time */
+/** Class used to set reference date/time for testing and convert input timeAvailableUntil to date timestamp */
 public final class Time {
 
-  private static Date date;
+  private Date date;
 
   /** Constructor */
   public Time() {
-    date = new Date();
+    this(new Date());
   }
 
   /** Constructor with manually set date */
-  public Time(Date date) {
+  private Time(Date date) {
     this.date = date;
   }
 
   /** Return today's date with time of hours:minutes in milliseconds since epoch */
-  public static long getTimeMillis(int hours, int minutes) {
-    
+  public long getTimeMillis(int hours, int minutes) {
+    // Validate hours and minutes
     if (hours < 0 || hours >= 24) {
       throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
     }
-
     if (minutes < 0 || minutes >= 60) {
       throw new IllegalArgumentException("Minutes can only be 0 through 59 (inclusive).");
     }
@@ -32,7 +31,7 @@ public final class Time {
     // Calculate today's date but with hours:minutes time
     // TODO: All times are currently today, wrap around times?
     Calendar c = Calendar.getInstance();
-		c.setTime(date);
+    c.setTime(date);
     c.set(Calendar.HOUR_OF_DAY, hours);
     c.set(Calendar.MINUTE, minutes);
     c.set(Calendar.SECOND, 0);
--- a/backend/src/main/java/com/google/sps/data/Time.java
+++ b/backend/src/main/java/com/google/sps/data/Time.java
@@ -19,7 +19,7 @@ public final class Time {
   }
 
   /** Return today's date with time of hours:minutes in milliseconds since epoch */
-  public static long getTimeMillis(int hours, int minutes) {
+  public static long getDateMillis(int hours, int minutes) {
     // Validate hours and minutes
     if (hours < 0 || hours >= 24) {
       throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
--- a/backend/src/main/java/com/google/sps/data/Time.java
+++ b/backend/src/main/java/com/google/sps/data/Time.java
@@ -14,7 +14,7 @@ public final class Time {
   }
 
   /** Constructor with manually set date */
-  private Time(Date date) {
+  public Time(Date date) {
     this.date = date;
   }
 
--- a/backend/src/main/java/com/google/sps/data/Time.java
+++ b/backend/src/main/java/com/google/sps/data/Time.java
@@ -6,7 +6,7 @@ import java.util.Calendar;
 /** Class used to set reference date/time for testing and convert input timeAvailableUntil to date timestamp */
 public final class Time {
 
-  private Date date;
+  private static Date date;
 
   /** Constructor */
   public Time() {
@@ -19,7 +19,7 @@ public final class Time {
   }
 
   /** Return today's date with time of hours:minutes in milliseconds since epoch */
-  public long getTimeMillis(int hours, int minutes) {
+  public static long getTimeMillis(int hours, int minutes) {
     // Validate hours and minutes
     if (hours < 0 || hours >= 24) {
       throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
--- a/backend/src/main/java/com/google/sps/data/Time.java
+++ b/backend/src/main/java/com/google/sps/data/Time.java
@@ -1,0 +1,42 @@
+package com.google.sps.data;
+
+import java.util.Date;
+import java.util.Calendar;
+
+/** Class to convert Time */
+public final class Time {
+
+  private static Date date;
+
+  /** Constructor */
+  public Time() {
+    date = new Date();
+  }
+
+  /** Constructor with manually set date */
+  public Time(Date date) {
+    this.date = date;
+  }
+
+  /** Return today's date with time of hours:minutes in milliseconds since epoch */
+  public static long getTimeMillis(int hours, int minutes) {
+    
+    if (hours < 0 || hours >= 24) {
+      throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
+    }
+
+    if (minutes < 0 || minutes >= 60) {
+      throw new IllegalArgumentException("Minutes can only be 0 through 59 (inclusive).");
+    }
+
+    // Calculate today's date but with hours:minutes time
+    // TODO: All times are currently today, wrap around times?
+    Calendar c = Calendar.getInstance();
+		c.setTime(date);
+    c.set(Calendar.HOUR_OF_DAY, hours);
+    c.set(Calendar.MINUTE, minutes);
+    c.set(Calendar.SECOND, 0);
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    if (minutes < 0 || minutes >= 60) {
+      throw new IllegalArgumentException("Minutes can only be 0 through 59 (inclusive).");
+    }
+
+    // Calculate today's date but with hours:minutes time
+    // TODO: All times are currently today, wrap around times?
+    Calendar c = Calendar.getInstance();
+		c.setTime(date);
```
dhruvi618, line 35: nit: indentation

backend/src/test/java/com/google/sps/TimeTest.java
```diff
@@ -0,0 +1,51 @@
...
+    Calendar c = Calendar.getInstance();
+    c.set(2020, 0, 1, 14, 0, 0);
+    c.set(Calendar.MILLISECOND, 0);
+    Date date = c.getTime();
+
+    time = new Time(date);
+  }
+
+  @Test
+  public void inputTimeToMillis() {
```
dhruvi618, line 29: Since the getTimeMillis() method has two edge cases that check for out of bound hours/minutes, would it be possible to add tests that ensure an IllegalArgumentException is thrown?

backend/src/test/java/com/google/sps/TimeTest.java
```diff
@@ -0,0 +1,51 @@
...
+/** */
+public final class TimeTest {
+
+  private Time time;
+
+  @Before
+  public void setUp() {
+    // Set "current" date to  1/1/2020 2:00pm GMT
+    Calendar c = Calendar.getInstance();
+    c.set(2020, 0, 1, 14, 0, 0);
```
dhruvi618, line 21: What do you think about using a date formatter to try to establish a common way to set dates and potentially catch any date errors? This is the documentation for the [SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) and I also found an [example](https://mkyong.com/java/java-convert-date-and-time-between-timezone/) of how it's used with the Calendar and other classes since I know in the future we might need to convert between timezones.

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
...
+  public Time() {
+    date = new Date();
+  }
+
+  /** Constructor with manually set date */
+  public Time(Date date) {
+    this.date = date;
+  }
+
+  /** Return today's date with time of hours:minutes in milliseconds since epoch */
```
dhruvi618, line 21: nit: I think the `/**` makes it a javadoc instead of a comment

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
+package com.google.sps.data;
+
+import java.util.Date;
+import java.util.Calendar;
+
+/** Class to convert Time */
+public final class Time {
+
+  private static Date date;
```
kmh287, line 9: why static?

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
...
+
+    // Calculate today's date but with hours:minutes time
+    // TODO: All times are currently today, wrap around times?
+    Calendar c = Calendar.getInstance();
+		c.setTime(date);
+    c.set(Calendar.HOUR_OF_DAY, hours);
+    c.set(Calendar.MINUTE, minutes);
+    c.set(Calendar.SECOND, 0);
+    c.set(Calendar.MILLISECOND, 0);
+    return c.getTime().getTime();
```
tcwanggoogle, line 40: (Just want to make sure it's not a typo) Do we really need to `getTime()` twice?

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
+package com.google.sps.data;
+
+import java.util.Date;
+import java.util.Calendar;
+
+/** Class to convert Time */
+public final class Time {
```
tcwanggoogle, line 7: (A question from someone who isn't familiar with Java) This class seems to implement some very common functionalities. Don't we have standard Java libraries to do similar things, or can we try to use standard library to avoid writing new functions if possible?

backend/src/test/java/com/google/sps/TimeTest.java
tcwanggoogle: +1 to add tests covering corner cases.

backend/src/test/java/com/google/sps/TimeTest.java
tcwanggoogle: +1 to encourage using standard libraries whenever possible

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: The main method (getTimeInMillis()) is pretty specific to our implementation. It takes an input "time" such as 11:00am and converts that to a Date timestamp on the current date but at that time. This implementation isn't final, though, because it all depends on what input we get from the front end input form.

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: The first getTime() of Calendar returns a Date object, and the second getTime() of Date returns a long timestamp.

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: I got rid of this static, made the getTimeMillis() method static, and changed the references in TimeTest.java

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private static Date date;
+
+  /** Constructor */
+  public Time() {
+    date = new Date();
+  }
```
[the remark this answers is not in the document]
jzwang8: Changed it so that this constructor called the other. Is it correct now?

backend/src/main/java/com/google/sps/data/Time.java
```diff
@@ -0,0 +1,42 @@
+package com.google.sps.data;
+
+import java.util.Date;
+import java.util.Calendar;
+
+/** Class to convert Time */
```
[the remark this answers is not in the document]
jzwang8: Explained. Let me know if it's still unclear. Basically, it's a class I'm using for both testing (setting a new reference date) and converting user input to an actual date.

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: I meant to have it be a javadoc. Should it not be?

backend/src/test/java/com/google/sps/TimeTest.java
jzwang8: Right now for the prototype, everything is in the long timestamp format, but for the MVP and depending on user input format, I'll make it a more standardized time format.

backend/src/test/java/com/google/sps/TimeTest.java
jzwang8: Added tests, thanks for the suggestion!

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: For some reason, that weird indentation doesn't show in my editor?

backend/src/main/java/com/google/sps/data/Time.java
tcwanggoogle: Ok. Interesting. I wonder why the first is not called `getDate`, which would make more sense. But that's not your fault

backend/src/main/java/com/google/sps/data/Time.java
tcwanggoogle: Hopefully with formatter in place, we won't see these problems in the future.

backend/src/main/java/com/google/sps/data/Time.java
tcwanggoogle: Cool. So maybe explain a bit about this special date + time combination in the javadoc? (The name "Time" makes me think it's kind of a standard interpretation of Time)

jzwang8: Changed the method name to getDateMillis() to clarify.

backend/src/main/java/com/google/sps/data/Time.java
jzwang8: Edit: I changed it back to being static for it to be referenced easily in the Test. For now, this method is only used in the tests, but I also wanted it to be used when setting timeAvailableUntil later.

backend/src/main/java/com/google/sps/data/Time.java
kmh287: I agree with Tiancong that "Time" might be too generic a name for this class. 

I'm wondering though if we could leverage the java 8 time libraries to simplify this somewhat. Could we implement similar functionality with:

```
long timestamp = LocalDate.now().atTime(hour, minute).toEpochSecond(ZoneOffset.UTC)
```
https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html

backend/src/test/java/com/google/sps/TimeTest.java
```diff
@@ -0,0 +1,71 @@
...
+    c.set(Calendar.MINUTE, 30);
+    Assert.assertEquals(c.getTime().getTime(), Time.getDateMillis(10, 30));
+
+    // 1/1/2020 11:45pm GMT
+    c.set(Calendar.HOUR_OF_DAY, 23);
+    c.set(Calendar.MINUTE, 45);
+    Assert.assertEquals(c.getTime().getTime(), Time.getDateMillis(23, 45));
+  }
+
+  @Test (expected = IllegalArgumentException.class)
```
kmh287, line 52: nice!

backend/src/test/java/com/google/sps/TimeTest.java
dhruvi618: Ok, sounds good

backend/src/main/java/com/google/sps/data/Time.java
dhruvi618: Nevermind, it should be. Sorry for the confusion.

PR #9, opened 2020-06-26
Add bare js fetch query

jzwang8:
This is untested--initial thoughts on how to connect with frontend.

### DIFF
--- a/frontend/src/match.js
+++ b/frontend/src/match.js
@@ -1,0 +1,29 @@
+// @Front-end: Call startMatch() on form submit
+/** Add participant, find match if possible, keep searching for match every 5 seconds */
+function startMatch() {
+  addParticipant();
+  searchMatch();
+}
+
+/** Add participant, find match with existing participants if possible */
+function addParticipant() {
+  const params = new URLSearchParams();
+  fetch('/add-participant', {method: 'POST', body: params});
+}
+
+/** Search for match in list of all matches */
+function searchMatch() {
+  fetch('/search-match').then(response => response.json()).then((match) => {
+    // No match yet
+    if (match.id == 0) {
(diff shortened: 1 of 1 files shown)


frontend/src/match.js
```diff
@@ -0,0 +1,29 @@
...
+function addParticipant() {
+  const params = new URLSearchParams();
+  fetch('/add-participant', {method: 'POST', body: params});
+}
+
+/** Search for match in list of all matches */
+function searchMatch() {
+  fetch('/search-match').then(response => response.json()).then((match) => {
+    // No match yet
+    if (match.id == 0) {
```
dhruvi618, line 18: What do you think about sending back a `matchStatus` which is true or false based on whether a match was found or not?

frontend/src/match.js
```diff
@@ -0,0 +1,29 @@
...
+      // Search every 5 seconds
+      setTimeout(searchMatch, 5000);
+    }
+    // Found match (eventually new page)
+    // @Front-end: ADD 'match-container' to CSS!
+    else {
+      const matchElement = document.getElementById('match-container');
+      matchLdapElement.innerText = match.firstldap + " " + match.secondLdap + " " + match.duration;
```
tcwanggoogle, line 26: nit: "l" and "L" in two variable names? Also, if renaming is not hard, I'd suggest we replace "ldap" with "username" for clarity.

dhruvi618:
Can we wait to merge this until I finalize the corresponding UI portion and learn more about the details relating to fetching and posting using React? I currently don't have the `match-container` element and am still working on getting the states and properties of all the form inputs

jzwang8:
Not pushing these changes to master

PR #15, opened 2020-07-01
Changed all Date/Calendar to ZonedDateTime, add startTimeAvailable to Participant, changed ldap to username

jzwang8:
Past PRs to refer to:
Add matching algorithm and tests
- FindMatchQuery.java.
- FindMatchQueryTest.java
Add servlets to query algorithm
- Match.java
- Participant.java
- AddParticipantServlet.java
- SearchMatchServlet.java
Add Time class and tests
- TimeHelper.java
- TimeHelperTest.java


backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -99,53 +110,75 @@ public void compatibleTime() {
...
+
     Match match = query.findMatch(Arrays.asList(participantA), participantB);
-    
+
     assertThat(match).isNull();
   }
 
   @Test
   public void threeParticipants13() {
     // Three participants, 1st and 2nd aren't compatible, but 1st and 3rd are
```
dhruvi618, line 127: nit: the tests have participant A, B, and C, but the comments and test names are in terms of 1, 2, and 3 (this is the same comment i have for the following tests)

dhruvi618:
LGTM!

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Ooh good catch! Thanks, done.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,42 +14,35 @@
...
-  private static int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
-  
+  private static int PADDING_TIME =
+      15; // extra padding time to ensure large enough meeting time block
+
   /** Constructor */
   public FindMatchQuery() {
-    clock = new Clock();
+    this(ZonedDateTime.now());
   }
```
kmh287, line 33: Can this constructor be removed in favor of the other one?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,42 +14,35 @@
...
 
   /** Constructor with manually set date */
-  public FindMatchQuery(Clock clock) {
-    this.clock = clock;
+  public FindMatchQuery(ZonedDateTime dateTime) {
```
kmh287, line 36: Please add javadocs to this method. Specifically as a user of this class I'd want to know what the preconditions are for the provided date time.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -58,23 +51,39 @@ public Match findMatch(List<Participant> participants, Participant newParticipan
...
 
       if (compatibleTime) {
         // TODO: change match ID (currently -1 for easy error checking)
-        return new Match(/* id= */ -1L, /* firstParticipant= */ newParticipant, /* secondParticipant= */ currParticipant, /* duration= */ duration, /* timestamp= */ date.getTime());
+        return new Match(
+            /* id= */ -1L,
+            /* firstParticipant= */ newParticipant,
+            /* secondParticipant= */ currParticipant,
+            /* duration= */ duration,
+            /* timestamp= */ dateTime.toInstant().toEpochMilli());
```
kmh287, line 74: non-blocking: Adding these to all arguments is entirely ok. Internally we only require them for literals like `-1L` since named variables often include details about the corresponding parameter e.g. `duration` above is passed as the argument for the `duration` parameter which is already unambiguous.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -58,23 +51,39 @@ public Match findMatch(List<Participant> participants, Participant newParticipan
...
     return null;
   }
+
+  /** Return earlier of two ZonedDateTime objects */
+  private ZonedDateTime getEarlier(ZonedDateTime first, ZonedDateTime second) {
+    if (first.isBefore(second)) {
+      return first;
+    } else {
+      return second;
+    }
```
kmh287, line 87: optional:

`return first.isBefore(second) ? first : second`

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -13,20 +13,24 @@
...
+  public Match(
+      long id,
+      Participant firstParticipant,
+      Participant secondParticipant,
+      int duration,
+      long timestamp) {
+    // Note: Match includes entire Participants instead of just usernames to distinguish between a
+    // user participating
+    // and finding a match multiple times. This wouldn't be an issue if we deleted matches from the
+    // datastore.
```
kmh287, line 27: This looks like it would be better suited as class javadocs. Also could you elaborate in the comment on why two usernames wouldn't be sufficient?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,27 +1,35 @@
...
   /** Initialize constructor fields */
-  public Participant(long id, String ldap, long timeAvailableUntil, String timezone, int duration, long timestamp) {
+  public Participant(
+      long id,
+      String username,
+      ZonedDateTime startTimeAvailable,
+      ZonedDateTime endTimeAvailable,
+      int duration,
+      long timestamp) {
     this.id = id;
```
kmh287, line 29: `startTimeAvailable` and `endTimeAvailable` have an implicit dependency on each other that `endTimeAvailable` is after `startTimeAvailable` but there are no guards to prevent this class from being constructed with date times that don't match this invariant. I would suggest either:

1. Replacing endTimeAvailable with an availability duration. I believe that a ZoneDateTime start and a long duration should be enough to construct a ZoneDateTime for the end. The main downside here is that this class has to worry about two separate durations, one for the availability window length and one for the desired meeting length
2. Add a precondition check (for example with this class https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html) that asserts that start is before after.

backend/src/main/java/com/google/sps/data/TimeHelper.java
```diff
@@ -1,42 +1,37 @@
...
+  /** Return today's date with time of hour:minute */
+  public static ZonedDateTime getNewTimeToday(int hour, int minute) {
+    // Validate hour and minute
+    if (hour < 0 || hour >= 24) {
       throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
     }
-    if (minutes < 0 || minutes >= 60) {
+    if (minute < 0 || minute >= 60) {
       throw new IllegalArgumentException("Minutes can only be 0 through 59 (inclusive).");
     }
```
kmh287, line 31: will .withHour() and .withMinute() throw in these circumstances? If so you might be able to just defer to those methods to throw the necessary exceptions.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -24,42 +24,60 @@
...
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Request parameter values
-    // TODO: Error check input values (before allowing to submit)
+    // TODO: Check input values are filled (before allowing to submit)
     UserService userService = UserServiceFactory.getUserService();
     String email = userService.getCurrentUser().getEmail();
-    String ldap = email.split("@")[0];
-    long timeAvailableUntil = convertToPositiveLong(request.getParameter("timeAvailableUntil"));
+    if (email == null) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Invalid email.");
```
kmh287, line 53: and then return?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -24,42 +24,60 @@
...
+                request.getParameter("endTimeAvailable"))); // TODO: figure out input format
+    LocalDateTime endTimeAvailableLocal =
+        endTimeAvailableInstant.atZone(ZoneId.of("UTC")).toLocalDateTime();
+    ZonedDateTime endTimeAvailable = endTimeAvailableLocal.atZone(zoneId);
+
     int duration = convertToPositiveInt(request.getParameter("duration"));
-    if (email == null || timeAvailableUntil == -1L || duration == -1) {
-      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Invalid input(s).");
+    if (duration == -1) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Invalid duration.");
```
kmh287, line 71: and return?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -34,20 +33,22 @@
...
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Get participant ldap
+    // Get participant username
     UserService userService = UserServiceFactory.getUserService();
     String email = userService.getCurrentUser().getEmail();
-    String ldap = email.split("@")[0];
+    if (email == null) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Invalid email.");
```
kmh287, line 46: return?

backend/src/test/java/com/google/sps/TimeHelperTest.java
```diff
@@ -0,0 +1,73 @@
...
+        ZonedDateTime.of(
+            /* year= */ 2020,
+            /* month= */ 1,
+            /* date= */ 1,
+            /* hour= */ 14,
+            /* minute= */ 0,
+            /* second= */ 0,
+            /* nanosecond= */ 0,
+            /* zone= */ ZoneId.of("US/Eastern"));
+    Assert.assertEquals(/* expected */ dateTime, /* actual */ TimeHelper.getNewTimeToday(14, 0));
```
kmh287, line 43: use Truth?

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Added a precondition check to the Participant constructor.

backend/src/main/java/com/google/sps/data/TimeHelper.java
jzwang8: Good point, they throw DateTimeExceptions. I removed the manual checks and changed the TimeHelperTest.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,42 +14,35 @@
...
 
-  private Clock clock;
+  private ZonedDateTime dateTime;
   private static int MAX_DURATION_DIFF = 15; // maximum difference in duration to be compatible
-  private static int PADDING_TIME = 15; // extra padding time to ensure large enough meeting time block
-  
+  private static int PADDING_TIME =
+      15; // extra padding time to ensure large enough meeting time block
+
   /** Constructor */
```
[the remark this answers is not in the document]
jzwang8: Added. (I addressed these in earlier commits after you reviewed but before you submitted)

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -13,20 +13,24 @@
...
   /** Constructor */
-  public Match(long id, Participant firstParticipant, Participant secondParticipant, int duration, long timestamp) {
+  public Match(
+      long id,
+      Participant firstParticipant,
+      Participant secondParticipant,
+      int duration,
+      long timestamp) {
+    // Note: Match includes entire Participants instead of just usernames to distinguish between a
+    // user participating
```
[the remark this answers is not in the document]
jzwang8: I think the Java formatter broke the line into two because the first was too long. Regardless, changed this to a class javadoc.

backend/src/main/java/com/google/sps/data/TimeHelper.java
```diff
@@ -1,42 +1,37 @@
...
-import java.util.Calendar;
-import java.time.LocalDateTime;
+import java.time.ZonedDateTime;
 
-/** Class used to set reference date/time for testing and convert input timeAvailableUntil to date timestamp */
+/**
+ * Class used to set reference date/time for testing and convert input timeAvailableUntil to date
+ * timestamp
+ */
 public final class TimeHelper {
```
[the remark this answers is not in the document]
jzwang8: I tried getting rid of the TimeHelper class and moving getNewTimeToday() to FindMatchQueryTest, but I was running into a lot of issues with ZonedDateTime being referenced from a non-static context that I couldn't fix so I decided to keep TimeHelper.

backend/src/test/java/com/google/sps/TimeHelperTest.java
```diff
@@ -40,33 +42,33 @@ public void inputTimeToMillis() {
...
 
     // 1/1/2020 10:30am ET
     dateTime = dateTime.withHour(10).withMinute(30);
-    Assert.assertEquals(dateTime, TimeHelper.getNewTimeToday(10, 30));
+    assertThat(dateTime).isEqualTo(TimeHelper.getNewTimeToday(10, 30));
 
     // 1/1/2020 11:45pm ET
     dateTime = dateTime.withHour(23).withMinute(45);
-    Assert.assertEquals(dateTime, TimeHelper.getNewTimeToday(23, 45));
+    assertThat(dateTime).isEqualTo(TimeHelper.getNewTimeToday(23, 45));
```
kmh287: thanks for converting to truth. Now it's clearer than actual and expected have been flipped. Putting the actual value (the one under test) and the expected value in the right place makes error logs easier to read. 

dateTime is the expected value whereas TimeHelper.getNEwTimeToday(23, 45) is the actual value.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -73,9 +93,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
       // Match not found, insert participant entity into datastore
       Entity participantEntity = new Entity("Participant");
-      participantEntity.setProperty("ldap", ldap);
-      participantEntity.setProperty("timeAvailableUntil", timeAvailableUntil);
-      participantEntity.setProperty("timezone", timezone);
+      participantEntity.setProperty("username", username);
+      participantEntity.setProperty("startTimeAvailable", startTimeAvailable);
+      participantEntity.setProperty("endTimeAvailable", endTimeAvailable);
       participantEntity.setProperty("duration", duration);
       participantEntity.setProperty("timestamp", timestamp);
```
kmh287, line 100: since each of these keys are referenced twice, there's a possibility to introduce bugs when one instance is changed and the other isn't. For example if "ldap" was updated here but not below. 

To guard against this, I'd recommend extracting string constants for each of these keys. Then when you want to change the key, it only has to be updated in a single place.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -56,36 +58,37 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        int duration = (int) entity.getProperty("duration"); 
-        long timestamp = (long) entity.getProperty("timestamp");
-        Match match = new Match(id, firstParticipant, secondParticipant, duration, timestamp);
-        matches.add(match);
+      long id = (long) entity.getKey().getId();
+      Participant firstParticipant = (Participant) entity.getProperty("firstParticipant");
+      Participant secondParticipant = (Participant) entity.getProperty("secondParticipant");
+      int duration = (int) entity.getProperty("duration");
+      long timestamp = (long) entity.getProperty("timestamp");
+      Match match = new Match(id, firstParticipant, secondParticipant, duration, timestamp);
```
kmh287, line 66: same comment about key Strings

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: What if I used newParticipant.getUsername(), .getStartTimeAvailable(), etc?

jzwang8: Done.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -37,6 +37,16 @@
 @WebServlet("/search-match")
 public class SearchMatchServlet extends HttpServlet {
 
+  /** Datastore Key constant string */
+  private static final String KEY_MATCH = "Match";
+
+  /** Datastore Property constant strings */
```
kmh287, line 43: tiny nit: this will show up in docs pages / editor docs as only the documentation for PROPERTY_FIRSTPARTICIPANT but not the other properties. This would be a case where you should use // instead of /** */ since this is a general comment and not a single item's javadocs

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -37,6 +37,16 @@
...
+  private static final String PROPERTY_FIRSTPARTICIPANT = "firstParticipant";
+
```
kmh287, line 45: tiny nit: delete newline and group all these together?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,42 +14,32 @@
...
-  /** Constructor */
-  public FindMatchQuery() {
-    clock = new Clock();
-  }
+  /** Reference date time for "current" time */
+  private ZonedDateTime dateTime;
+  /** Maximum difference in duration to be compatible */
+  private static int MAX_DURATION_DIFF = 15;
+  /** extra padding time to ensure large enough meeting time block */
+  private static int PADDING_TIME = 15;
```
kmh287, line 30: static constants should be declared at the top of the class, before any instance fields

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,42 +14,32 @@
...
 
-  /** Constructor with manually set date */
-  public FindMatchQuery(Clock clock) {
-    this.clock = clock;
+  /** Set "current" date and time to calculate whether a match is possible in time */
```
kmh287, line 32: Sorry to keep harping on this but I still think we can find a better name for the field and parameter here. "Current" time doesn't work because it will be stale immediately after instantiation. For example if this class were instantiated, kept alive for 24 hours, and then used to perform a match, it would produce incorrect results because this field is inaccurate. 

The right approach here would be to get the current time `ZoneDateTime` within the findMatch method so that it's always accurate. You can use ZoneDateTime.now(Clock) to get the current time within `findMatch` while also constructing this class with a Clock (as before) to facilitate testing.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -34,58 +33,72 @@
...
-      if (ldap.equals(firstParticipantLdap) || ldap.equals(secondParticipantLdap)) {
+      String firstParticipantUsername = match.getFirstParticipant().getUsername();
+      String secondParticipantUsername = match.getSecondParticipant().getUsername();
+      if (username.equals(firstParticipantUsername) || username.equals(secondParticipantUsername)) {
         JSONObject matchExists = new JSONObject();
         matchExists.put("matchStatus", "true");
-        matchExists.put("firstParticipantLdap", firstParticipantLdap);
-        matchExists.put("secondParticipantLdap", secondParticipantLdap);
+        matchExists.put("firstParticipantUsername", firstParticipantUsername);
+        matchExists.put("secondParticipantUsername", secondParticipantUsername);
```
kmh287, line 92: constants

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Since I instantiate this class every time a new participant is added, the "current" time is the time the participant is added. Should I make the FindMeetingQuery class static so I don't have to instantiate it every time? Or should I just keep it the way it is now and do the ZonedDateTime.now(Clock) thing?

kmh287: Clock sounds like the right way to go. The code instantiates the class every time a new participant is added 'now' but it's possible someone could come along later and try to reuse the class multiple times even though currently it only works correctly immediately after instantiation. Changing to a clock would allow using a single instance for multiple runs of findMatch.

backend/src/main/java/com/google/sps/data/TimeHelper.java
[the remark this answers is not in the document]
tcwanggoogle: Sure, but keep in mind to fix this in future PRs, because when we need to add new features, we should stick with one class to expand, so it's better to get rid of TimeHelper in the long run.

backend/src/main/java/com/google/sps/data/TimeHelper.java
kmh287: Thanks!

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -94,7 +95,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
     // Find immediate match if possible
-    FindMatchQuery query = new FindMatchQuery(ZonedDateTime.now());
+    ZonedDateTime dateTime = ZonedDateTime.now();
+    Clock clock = Clock.fixed(dateTime.toInstant(), dateTime.getZone());
```
kmh287: Using a fixed clock still leaves the potential for inaccurate results from findMatch if the FindMatchQuery object exists for a long time. The key benefit of passing a clock is so that findMatch can be called at multiple times and always use the correct time when running the algorithm. It looks like the Clock.systemDefaultZone() or better Clock.systemUTC() will work here.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Ah I see, thanks for explaining. Is it correct now? I used a fixed clock in the match query test, and in AddParticipantServlet, I passed Clock.systemUTC() to FindMatchQuery(), and instantiated ZonedDateTime in the findMatchQuery() method.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -69,8 +70,9 @@ public void setUp() {
             /* second= */ 0,
             /* nanosecond= */ 0,
             /* zone= */ ZoneId.of("US/Eastern"));
+    Clock clock = Clock.fixed(dateTime.toInstant(), dateTime.getZone());
```
kmh287, line 73: Nice! Glad to see the switch for the tests was straightforward.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Yep, I immediately realized my mistake right after pushing so I unfixed the clock in my "Unfix clock" commit above!

kmh287: Nice. Can't argue with a -4 minute turnaround on a PR comment

PR #18, opened 2020-07-07
Converting internal API to use Gmail instead of Twillio	

TcgVanguardTroll:
This PR will introduce the implementation of an internal Gmail Mailing system instead of Twilio. Due to the notification that third parties API's should be leaned away from as well as the ramifications of long term usage of the Twilio API given its constraints for a not for profit project we have decided to use the Gmail API.

The main changes in this PR include:

- The Main EmailNotifier class and its API which implements the `notifyuser` inner functionality in order to email the user inputted from within its Parameter 

The Files Impacted by this PR : 

- `.gitignore` : Adding various files to be ignored / don't need to be known on a public repo.

- `pom.xml`:  Adding more dependencies that  this project requires .

- `EmailNotifier.java`: Main file that this PR impacts, this file will include the majority of the Gmail API implementation.

- `TestEmailNotifier.java`: Test for Email Notifier, tests currently ensure that the emails and messages have the correct information as well as internal method calls have the correct behavior


backend/pom.xml
```diff
@@ -1,97 +0,0 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
```
kmh287, line 1: Accidental deletion?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+        Message message = new Message();
+        message.setRaw(encodedEmail);
+        return message;
+    }
+
+    public static void main(String[] args)
+            throws GeneralSecurityException, MessagingException, IOException {
+        EmailNotifier emailNotifier = new EmailNotifier("jordan", "grantjustice@google.com");
+        emailNotifier.notifyUser();
+    }
```
kmh287, line 142: delete?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Getter function that returns the string representing the email recipients name .
+     *
+     * @return String representing the name of the email recipient.
+     */
+    @Override
+    public String getName() {
+        return super.getName();
+    }
```
kmh287, line 152: no need to override if the implementation only defers to super

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+    public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
+
+        MimeMessage email =
+                createEmail(
+                        getToEmail(),
+                        "Ad-Lib Meeting Found",
+                        " Hey "
+                                + this.getName()
+                                + " Please Join your Ad-Lib meeting via the link below : \n"
+                                + " http://meet.google.com/new");
```
kmh287, line 165: TODO to include the real link?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+        final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
kmh287, line 166: tiny tiny nit: There's no need to mark this as final. The compiler can tell which variables are "effectively final"

backend/src/main/java/com/google/sps/notifs/Notifier.java
```diff
@@ -0,0 +1,45 @@
...
+package main.java.com.google.sps.notifs;
+
+import com.google.api.services.gmail.Gmail;
+
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+public abstract class Notifier {
```
kmh287, line 24: What other implementations did you have in mind for Notifier besides the gmail notifier? I'm still unsure if this should be an abstract class or if all the functionality should be merged into the gmail notifier.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+     * Global instance of the scopes required by this quickstart. If modifying these scopes, delete
+     * your previously saved tokens/ folder.
+     */
+    private static final List<String> SCOPES =
+            Collections.singletonList(GmailScopes.MAIL_GOOGLE_COM);
+
+    private static final String CREDENTIALS_FILE_PATH = "credentials.json";
+
+    /** Variable representing the email to be notified */
+    protected String toEmail;
```
kmh287, line 58: private final?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+        return email;
+    }
+
+    /**
+     * Create a message from an email.
+     *
+     * @param emailContent Email to be set to raw of message
+     * @return a message containing a base64url encoded email
+     * @throws IOException
+     * @throws MessagingException
```
kmh287, line 125: when are these exceptions thrown?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+import java.io.*;
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+
+/** Class representing the notification system capable of sending mails to the user. */
+public class EmailNotifier extends Notifier {
+
+    private static final String APPLICATION_NAME = "Ad-libs";
```
jzwang8, line 45: *Ad-lib

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final String CREDENTIALS_FILE_PATH = "credentials.json";
+
+    /** Variable representing the email to be notified */
+    protected String toEmail;
+
+    /**
+     * @param name Name of the recipient of the user
+     * @param to Email that this email is going to sent to.
+     */
+    public EmailNotifier(String name, String to) {
```
jzwang8, line 64: Change "to" to "toEmail" and "name" to "recipientName"?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Create a MimeMessage using the parameters provided.
+     *
+     * @param to email address of the receiver
+     * @param subject subject of the email
+     * @param bodyText body text of the email
+     * @return the MimeMessage to be used to send email
+     * @throws MessagingException
+     */
+    public static MimeMessage createEmail(String to, String subject, String bodyText)
```
jzwang8, line 104: "toEmail"?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+            throws MessagingException {
+
+        Properties props = new Properties();
+        Session session = Session.getDefaultInstance(props, null);
+
+        MimeMessage email = new MimeMessage(session);
+
+        email.setFrom(new InternetAddress(to));
```
jzwang8, line 112: Hmm should setFrom() have the "to" email as an argument?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
jzwang8, line 104: Do you need a "fromEmail"?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
jzwang8: Oh I see, for testing, you set the from and to emails to your email. Maybe make sure to keep "to" and "from" separate and just set each as your email to test.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
jzwang8: +1 maybe add a Test class

backend/src/main/java/com/google/sps/notifs/Notifier.java
jzwang8: +1 I think we can just use the Gmail API for emails (no texts, etc.)

pom.xml
```diff
@@ -0,0 +1,118 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
jzwang8, line 1: Should this pom.xml file be the same as the one I'm working with in backend/ ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main.java.com.google.sps.notifs;
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver;
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeFlow;
+import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp;
+import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
```
dhruvi618, line 21: tiny nit: alphabetize import statements

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+        this.toEmail = to;
+    }
+
+    /**
+     * Creates an authorized Credential object.
+     *
+     * @param HTTP_TRANSPORT The network HTTP Transport.
+     * @return An authorized Credential object.
+     * @throws IOException If the credentials.json file cannot be found.
+     */
```
dhruvi618, line 75: I really like these comments!

pom.xml
dhruvi618: +1

dhruvi618:
Quick overall comment about indentation: It seems like the indentation size your editor is defaulting to is 4 spaces rather than 2 spaces. To be consistent with indentation in the entire repo, I think we should all use 2 spaces for indentation. 

I found this on one of the style guides: https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Alphabetized

pom.xml
TcgVanguardTroll: I think we're all going to be using the same pom.xml, I had moved it previously just to test my maven. If the files are override and some dependencies go missing i think we can ultimately resolve them

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Because we are going to be sending from the same dummy account int he final version I was thinking of just passing in the value directly rather than using it as a parameter

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Currently, only authenticated accounts can send emails and currently I'm the only authenticated sender so whichever parameter that is passed will result in me sending the email but ideally I want to use this parameter to send a dummy email

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Renamed

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Renamed

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Fixed

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Updated comment

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Done

backend/src/main/java/com/google/sps/notifs/Notifier.java
TcgVanguardTroll: I was thinking of adding text messages and potentially more but given the scope of the api's that we are limited to I think this is a good suggestion

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Deleted, and I'm currently looking for ways to automatically test this API functionality

backend/pom.xml
TcgVanguardTroll: Had moved it so that I could call maven in IntelliJ without reconfiguring due to me not knowing how

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Most of them come from the documentation, but they have been really insightful for moving forward in documenting code

backend/pom.xml
kmh287: But your teammates are using this file in its current location. It can move if doing so is required technically or is better organizationally but I wouldn't move it just to make intelliJ happy. If there's no way to configure intelliJ to use the file in its current location then there's always the `mvn` command line interface. WDYT?

FYI @jzwang8

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,174 @@
...
+    String encodedEmail = Base64.encodeBase64URLSafeString(bytes);
+    Message message = new Message();
+    message.setRaw(encodedEmail);
+    return message;
+  }
+
+  /** Getter function that returns the string representing the email recipients name . */
+  public String getRecipientName() {
+    return recipientName;
+  }
```
kmh287, line 146: are the getters for recipient name and email ever used or do you foresee a future use case? If not we can remove them

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,174 @@
...
+    InputStream in = new FileInputStream(CREDENTIALS_FILE_PATH);
+    GoogleClientSecrets clientSecrets =
+            GoogleClientSecrets.load(JSON_FACTORY, new InputStreamReader(in));
+
+    // Build flow and trigger user authorization request.
+    GoogleAuthorizationCodeFlow flow =
+            new GoogleAuthorizationCodeFlow.Builder(
+                    HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES)
+                    .setDataStoreFactory(
+                            new FileDataStoreFactory(new java.io.File(TOKENS_DIRECTORY_PATH)))
```
kmh287, line 93: nit: `File` instead of `java.io.File`

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,174 @@
...
+   * @param subject subject of the email
+   * @param bodyText body text of the email
+   * @return the MimeMessage to be used to send email
+   * @throws MessagingException for other failures
+   */
+  public static MimeMessage createEmail(String toEmail, String subject, String bodyText)
+          throws MessagingException {
+
+    Properties props = new Properties();
+    Session session = Session.getDefaultInstance(props, null);
```
kmh287, line 113: /* param= */ for the null argument

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,174 @@
...
+
+    MimeMessage email = new MimeMessage(session);
+
+    email.setFrom(new InternetAddress("Adlib-Step@gmail.com"));
+    email.addRecipient(javax.mail.Message.RecipientType.TO, new InternetAddress(toEmail));
```
kmh287, line 118: Can RecipientType be imported to avoid the fully-qualified name here?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
jzwang8: I see. Still, I think it's important to distinguish between the from and to emails even if they are the same thing right now.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
jzwang8: Sounds good. Should probably set that as private static final String constant above instead of hardcoding it into this method.

backend/pom.xml
jzwang8: Agreed, I'm using the backend/pom.xml file and will push changes to it once my PR gets approved, so I think it would be best to keep this file!

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: I was mainly using them in the body text of the email as well as generally in using the create email function would you suggest directly accessing them the email and name or using getters ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Imported

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Added the parameter comment of authenticator just to be more precise .

backend/pom.xml
TcgVanguardTroll: Yes, that is why is reverted the deletion after

TcgVanguardTroll: The MVN command interface wouldn't work due to here not being a POM.xml in the directory in which my project is structured and I haven't fully configured intellij on my corporate laptop . The mvn cli is the first thing that i tried and whenever i would run it i would get `The goal you specified requires a project to execute but there is no POM in this directory (/home/grantjustice/IdeaProjects/Ad-lib). Please verify you invoked Maven from the correct directory`

kmh287: what if you cd into the backend directory and re-run the command?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+  private static final String APPLICATION_EMAIL = "Adlib-Step@gmail.com";
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  /**
+   * Global instance of the scopes required by this quickstart. If modifying these scopes, delete
+   * your previously saved tokens/ folder.
+   */
+  private static final List<String> SCOPES =
+          Collections.singletonList(GmailScopes.MAIL_GOOGLE_COM);
```
kmh287, line 60: ImmutableList.of() is slightly shorter and is explicitly immutable

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
kmh287: It definitely makes sense to use the getters especially if you're concerned that they will be refactored to be "non-pure" i.e. to do more than just returning a field directly. I could go either way but if we keep the getters, they should be private until we have a use-case to access them from outside the class

backend/pom.xml
```diff
@@ -1,97 +1,118 @@
...
```
kmh287, line 1: Looks like the entire file was reformatted. Did that happen automatically with one of the auto-formaters? Were there any other changes to this file or just addition / removal of blank space?

backend/pom.xml
TcgVanguardTroll: I get the same error

backend/pom.xml
TcgVanguardTroll: I might've used intellij's auto-format when i had reverted the file which caused the spacing changes but in terms of additions i did add  additional dependencies which were just the dependencies needed for Gmail such as oath client etc.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/pom.xml
kmh287: Can you share the error?

TcgVanguardTroll: [ERROR] The goal you specified requires a project to execute but there is no POM in this directory (/home/grantjustice/IdeaProjects/Ad-lib). Please verify you invoked Maven from the correct directory. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MissingProjectException

kmh287: That still looks like it's being run from the root instead of the backend directory. Can you try

```
cd /home/grantjustice/IdeaProjects/Ad-lib/backend
mvn compile
```

TcgVanguardTroll:
> It looks like there are still two conflicts. Make sure you rebase the changes on top of master again, otherwise the Java linter might not work in this PR.

I'm not sure if this is because I incorrectly re-based but my Java Format is failing

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -14,8 +14,8 @@
 
 package com.google.sps;
 
-import com.google.sps.data.Match;
-import com.google.sps.data.Participant;
+import main.java.com.google.sps.data.Match;
+import main.java.com.google.sps.data.Participant;
```
kmh287, line 18: here and elsewhere: please make sure all the package names remain as com.google.sps....

TcgVanguardTroll: On IntelliJ  it states that package names should correspond to the file path, in that case should all work be in a com.google.sps. package ?

TcgVanguardTroll: Or is this only for Janet's Matching algorithm files ?

kmh287: yeah, since this file is in sps the package is com.google.sps

a file in java/com/google/sps/data would have package com.google.sps.data

TcgVanguardTroll: Okay just to clarify, Currently our backend files are in `src/main/java/com/google/sps/ `folder name``  . I believe the issue is storing the files in the main folder instead of directly in src so a file in src/main/java/com/google/sps/data should be moved from this directory into a `src/com/google/sps/data` ?

TcgVanguardTroll: Or should I be ignoring IntelliJ ?

kmh287: Ignore intelliJ.

TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -11,9 +11,9 @@
   /** Datastore entity ID */
   private final long id;
   /** First participant */
-  private final Participant firstParticipant;
+  private final com.google.sps.data.Participant firstParticipant;
```
tcwanggoogle, line 14: `import com.google.sps.data.Participant` so we don't need the full name all the time?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps.servlets;
+package main.java.com.google.sps.servlets;
```
tcwanggoogle, line 15: More package path to fix?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps;
+package test.java.com.google.sps;
```
tcwanggoogle, line 15: More package path to fix?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -12,10 +12,11 @@
...
 
 import static com.google.common.truth.Truth.assertThat;
 
+import com.google.sps.FindMatchQuery;
```
tcwanggoogle, line 19: Why do we need to fix this import in this PR?

tcwanggoogle:
Also, please add google-java-format-1.8-all-deps.jar to .gitignore

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Fixed

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
TcgVanguardTroll: Fixed

backend/src/main/java/com/google/sps/data/Match.java
TcgVanguardTroll: I cannot import the Class due to the how the packages are defined causing the import to go unused

kmh287: If this will compile as-is then import com.google.sps.data.Participant should work too

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
TcgVanguardTroll: The naming was wrongly changed, I was just correcting it

backend/src/main/java/com/google/sps/data/Match.java
TcgVanguardTroll: Whenever i import com.google.sps.data.Participant the linter fails due to unused import staements

kmh287: Ah, I see why. Participant is also in this package so it can be referenced just as `Participant` without an import statement.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,175 @@
...
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  /**
+   * Global instance of the scopes required by this quickstart. If modifying these scopes, delete
+   * your previously saved tokens/ folder.
+   */
+  private static final List<String> SCOPES = ImmutableList.of(GmailScopes.MAIL_GOOGLE_COM);
+
+  private static final String CREDENTIALS_FILE_PATH = "credentials.json";
+
+  /** Variable representing the email to be notified */
```
kmh287, line 63: /** The email to be notified */

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,175 @@
...
+  private final String toEmail;
+
+  private final String recipientName;
+
+  /**
+   * @param recipientName Name of the recipient of the user
+   * @param toEmail Email that this email is going to sent to.
```
kmh287, line 70: two very tiny nits: 'Email address' and 'going to be sent to'

frontend/yarn.lock
```diff
@@ -11421,3 +11421,4 @@ yargs@^13.3.0:
     which-module "^2.0.0"
     y18n "^4.0.0"
     yargs-parser "^13.1.2"
+© 2020 GitHub, Inc.
```
kmh287, line 11424: revert

credentials.json
```diff
@@ -0,0 +1,10 @@
+{
+  "web": {
+    "client_id": "294321754695-t48b7ubomafqcitnmve2ngeotg4npsd2.apps.googleusercontent.com",
+    "project_id": "google.com:spontaneously",
+    "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+    "token_uri": "https://oauth2.googleapis.com/token",
+    "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+    "client_secret": "ByF-DdMdKvO-jby_3TfADwDx"
```
kmh287, line 8: Now that this file is in this PR this is no longer a secret. A malicious person could find this PR and use this key to pretend to be the Ad-Lib project. Please remove this file from the PR and **generate a new secret**. Some useful info:
https://cloud.google.com/docs/authentication/api-keys#securing_an_api_key

backend/src/main/java/com/google/sps/data/Match.java
TcgVanguardTroll: Acknowledged

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Acknowledged

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -37,11 +37,11 @@ public long getId() {
     return id;
   }
 
-  public Participant getFirstParticipant() {
+  public com.google.sps.data.Participant getFirstParticipant() {
```
tcwanggoogle, line 40: Remove "com.google.sps.data." here and below at line 44

frontend/yarn.lock
TcgVanguardTroll: Reverted

credentials.json
TcgVanguardTroll: I have added credentials.json to gitignore

backend/src/main/java/com/google/sps/data/Match.java
TcgVanguardTroll: Ack

tcwanggoogle:
Looks like all my comments are addressed, except "add google-java-format-1.8-all-deps.jar to .gitignore and remove it from this PR". Will LGTM after that's fixed

TcgVanguardTroll:
> Since it looks like these files weren't meant as part of this PR, please revert:
> 
> * AddParticipantServlet.java
> * dispatch.yaml
> * yarn.lock
> 
> Could you also describe what the StoredCredential file is?

I have removed Stored Credential but if i am correct it is used to act as an authorization for using the API and it is created on each call

tcwanggoogle:
> > Since it looks like these files weren't meant as part of this PR, please revert:
> > 
> > * AddParticipantServlet.java
> > * dispatch.yaml
> > * yarn.lock
> > 
> > Could you also describe what the StoredCredential file is?
> 
> I have removed Stored Credential but if i am correct it is used to act as an authorization for using the API and it is created on each call

I am not sure if we should remove it. How did you get this file? Does it get auto-generated when we compile the backend code? If not and/or later we find this file is needed, we should add it back.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Create a MimeMessage using the parameters provided.
+   *
+   * @param toEmail email address of the receiver
+   * @param subject subject of the email
+   * @param bodyText body text of the email
+   * @return the MimeMessage to be used to send email
+   * @throws MessagingException for other failures
```
kmh287, line 109: Can you expand on the throws doc here. What failures are possible?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /** Getter function that returns the string representing the email recipients name . */
+  private String getRecipientName() {
+    return recipientName;
+  }
+
+  /** Function that access its api and using it sends an email */
+  //    TODO(): Create a dummy email for ad lib itself to send emails.
+  //    TODO(): Replace body to send real link to user instead of generic.
```
kmh287, line 152: please add issue numbers

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: There are a lot of other failures , I don't know how to encapsulate them as the are all referred to here : https://docs.oracle.com/javaee/7/api/javax/mail/class-use/MessagingException.html

kmh287: Which calls in this method can throw this exception? Under what circumstances will they throw?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Done

TcgVanguardTroll:
> > > Since it looks like these files weren't meant as part of this PR, please revert:
> > > 
> > > * AddParticipantServlet.java
> > > * dispatch.yaml
> > > * yarn.lock
> > > 
> > > Could you also describe what the StoredCredential file is?
> > 
> > 
> > I have removed Stored Credential but if i am correct it is used to act as an authorization for using the API and it is created on each call
> 
> I am not sure if we should remove it. How did you get this file? Does it get auto-generated when we compile the backend code? If not and/or later we find this file is needed, we should add it back.

I didn't remove the file itself ,just remove it from the PR the stored credential does get auto-generated each call and is used very often .

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -28,21 +27,16 @@
...
-import java.util.Properties;
+import com.google.inject.Inject;
+import org.apache.commons.codec.binary.Base64;
 import javax.mail.Message.RecipientType;
 import javax.mail.MessagingException;
 import javax.mail.Session;
 import javax.mail.internet.InternetAddress;
 import javax.mail.internet.MimeMessage;
-import org.apache.commons.codec.binary.Base64;
+import java.io.*;
```
kmh287, line 37: Please use granular imports instead of * imports. Star imports can add a lot of unexpected classes to the namespace and obscure which classes are actually needed.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -58,21 +52,30 @@
    */
   private static final List<String> SCOPES = ImmutableList.of(GmailScopes.MAIL_GOOGLE_COM);
 
-  private static final String CREDENTIALS_FILE_PATH = "credentials.json";
+  private static final String CREDENTIALS_FILE_PATH =
+      "/home/grantjustice/IdeaProjects/Ad-lib/backend/credentials.json";
```
kmh287, line 56: This will need to be a path that's accessible when the project is deployed.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -58,21 +52,30 @@
...
    * @param recipientName Name of the recipient of the user
-   * @param toEmail The email addresse that this email is going to be sent to.
+   * @param toEmail The email address that this email is going to be sent to.
    */
-  public EmailNotifier(String recipientName, String toEmail) {
+  @Inject
+  public EmailNotifier(String recipientName, String toEmail, Gmail gmail) {
+    //    if (gmail == null) {
+    //      throw new InvalidParameterException("Gmail Cannot be Null");
+    //    }
```
kmh287, line 75: delete unused? I think it's ok for this class to throw NPEs if gmail is invalid because a caller shouldn't be passing null to the constructor here.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -123,36 +145,11 @@ public static MimeMessage createEmail(String toEmail, String subject, String bod
...
-    return recipientName;
-  }
-
   /** Function that access its api and using it sends an email */
-  //    TODO(): Create a dummy email for ad lib itself to send emails.
-  //    TODO(): Replace body to send real link to user instead of generic.
-
-  public void notifyUser() throws MessagingException, GeneralSecurityException, IOException {
+  //    TODO(#35): Create a dummy email for ad lib itself to send emails.
+  //    TODO(#36): Replace body to send real link to user instead of generic.
```
kmh287, line 151: tiny nit: there's an extra blank line here

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -161,16 +158,21 @@ public void notifyUser() throws MessagingException, GeneralSecurityException, IO
...
                 + " Please Join your Ad-Lib meeting via the link below : \n"
                 + " http://meet.google.com/new");
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
-    Gmail service =
-        new Gmail.Builder(HTTP_TRANSPORT, JSON_FACTORY, getCredentials(HTTP_TRANSPORT))
-            .setApplicationName(APPLICATION_NAME)
-            .build();
     Message messageWithEmail = createMessageWithEmail(email);
-    service.users().messages().send("me", messageWithEmail).execute();
+    getService().users().messages().send("me", messageWithEmail).execute();
```
kmh287, line 162: Any benefit to using a getter instead of directly referencing `service` here?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -161,16 +158,21 @@ public void notifyUser() throws MessagingException, GeneralSecurityException, IO
...
+
+  /** Getter function that returns the string representing the email recipients name . */
+  public String getRecipientName() {
+    return recipientName;
+  }
+
+  /** Getter function that returns the Gmail Service that emailNotifier depends on . */
+  public Gmail getService() {
+    return service;
+  }
```
kmh287: Will any of these methods be used by production code? I agree with the intention here to make the class easier to test but we need to balance that against making too much of the class' internals visible for testing. The ideal is for the class to be fully testable through its public API that the non-test code uses.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -14,18 +14,17 @@
...
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import java.time.Clock;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import static com.google.common.truth.Truth.assertThat;
 
 @RunWith(JUnit4.class)
```
kmh287: Revert changes to this file?

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.sps.notifs.EmailNotifier;
+import org.junit.Assert;
+import org.junit.Test;
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+import java.io.IOException;
+import java.util.Properties;
+import static org.junit.Assert.*;
```
kmh287, line 12: Please switch over to [Truth](www.truth.dev). It's used extensively for Java tests at google and is already used in the other BE code tests.

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+  public MimeMessage createTestMessage() {
+    Properties properties = System.getProperties();
+    Session session = Session.getDefaultInstance(properties);
+    return new MimeMessage(session);
+  }
+
+  private String getTextFromMessage(MimeMessage message) throws MessagingException, IOException {
+    String result = "";
+    result = message.getContent().toString();
+    return result;
```
kmh287, line 25: return message.getContent().toString();

?

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  public void constructorShouldSetNameAndEmailTo() {
+    EmailNotifier emailNotifier = new EmailNotifier("John", "jdoe@gmail.com", null);
+    assertEquals("John", emailNotifier.getRecipientName());
+    assertEquals("jdoe@gmail.com", emailNotifier.getToEmail());
```
kmh287, line 32: Continuing from my comment before: 100% agree that we should test that the email addresses specified int he constructor are used. One way we could do that is test that the emails generated from the class have the right to and from email addresses. Then we can take out these two methods that expose internals of the class.

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  public void testMessageHasCorrectApplicationName() throws MessagingException {
+    EmailNotifier emailNotifier = new EmailNotifier("Jordan", "jjgrant@buffalo.edu", null);
+    MimeMessage mimeMessage = emailNotifier.createEmail("", "", "");
+    MimeMessage message = createTestMessage();
```
kmh287, line 39: Similar to my concern form earlier comments: this test relies on exposing the internals of EmailNotifier. In other words, if not for this test, would createEmail need to be a non-private method? 

What I'd recommend doing instead is mocking out the Gmail service, capturing the sent email, and asserting that its from address is equal to "Adlib-Step@gmail.com"

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+  public void testMessageShouldHaveCorrectBodyText() throws MessagingException, IOException {
+    EmailNotifier emailNotifier = new EmailNotifier("Jordan", "jjgrant@buffalo.edu", null);
+    MimeMessage mimeMessage = emailNotifier.createEmail("", "", "Random Text");
+    MimeMessage message = createTestMessage();
+    message.setText("Random Text");
+    Assert.assertEquals(
+        "These two strings should not be the same ",
+        getTextFromMessage(message),
+        getTextFromMessage(mimeMessage));
+  }
```
kmh287, line 67: same suggestion as the previous test method: As much as possible we want to test the intended public API of this class. We should be able capture the emails sent by the mock gmail service and assert that the subject and body are the same

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+import java.io.IOException;
+import java.util.Properties;
+import static org.junit.Assert.*;
+
+public class TestEmailNotifier {
+
```
kmh287, line 15: Now that the `Gmail` dependency can be injected through the constructor, this test can provide a mock Gmail instance that can be used to stub / verify interactions with `Gmail`. 

To start, I'd recommend adding a set up method to the test to setup the mock. Once the mock is setup, you can construct an EmailNotifier and make it a field on the test class. (The test class is recreated for each test method so there's no need to worry about state carrying over from one test to the next)

```
private EmailNotifier emailNotifier;

@Before
public void setUp() {
  Gmail gmail = mock(Gmail.class)
   ....

  emailNotifier = new EmailNotifier(...)
}
```

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.Test;
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+import java.io.IOException;
+import java.util.Properties;
+import static org.junit.Assert.*;
+
+public class TestEmailNotifier {
```
kmh287, line 14: tiny nit: the file and class name should be EmailNotifierTest. The convention for test classes is to add "Test" on the end of the unit being tested. Confusingly, by convention a TestFoo is a version of the Foo class that's meant as a helper for testing.

backend/src/test/java/com/google/sps/TestEmailNotifier.java
```diff
@@ -0,0 +1,109 @@
+package com.google.sps;
```
kmh287, line 1: Really great testing scenarios in this class. 👍

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
tcwanggoogle: +1 If we need a credential file to send email, maybe it should be inside this repo?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -22,6 +22,7 @@
 import java.time.ZoneId;
```
tcwanggoogle, line 22: Do we need to change this test in this PR? I think not, so maybe we should not touch this file.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Changed path

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Deleted

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: In the function we were setting service as the gmail Builder,because of your feedback I thought that having a getter would be the way to access all of our gmail api calls

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,181 @@
...
+  private final String recipientName;
+
+  /** The gmail dependency */
+  private final Gmail service;
+
+  /**
+   * @param recipientName Name of the recipient of the user
+   * @param toEmail The email address that this email is going to be sent to.
+   */
+  public EmailNotifier(String recipientName, String toEmail, Gmail gmail) {
```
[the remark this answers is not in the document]
TcgVanguardTroll: @kmh287

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
TcgVanguardTroll: Fixed

backend/src/test/java/com/google/sps/TestEmailNotifier.java
TcgVanguardTroll: Done !

backend/src/test/java/com/google/sps/TestEmailNotifier.java
TcgVanguardTroll: Noted

backend/src/test/java/com/google/sps/TestEmailNotifier.java
TcgVanguardTroll: I am confused because send takes a message but form what I've seen the extraction methods (get) are most within the MimeMessage class ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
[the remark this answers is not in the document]
kmh287: The issue with notifyUser is that it creates a dependency, the Gmail object, internally. As a consequence, swapping that dependency to a testing fake (or any other implementation) is not possible. By injecting that dependency here in the constructor, we allow the tests and production code to provide whichever implementation of `Gmail` is best suited for their purposes.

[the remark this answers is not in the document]
tcwanggoogle: Thanks for the implementation. SGTM

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
tcwanggoogle: Forgot to push the change? It's still here

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Create a MimeMessage using the parameters provided.
+   *
+   * @param toEmail email address of the receiver
+   * @param subject subject of the email
+   * @param bodyText body text of the email
+   * @return the MimeMessage to be used to send email
+   * @throws MessagingException if there was a problem accessing the Store
+   */
+  public MimeMessage createEmail(String toEmail, String subject, String bodyText)
```
tcwanggoogle, line 133: Since it's not a static function anymore, there's no need to pass `toEmail`, as it's a property of the object.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,182 @@
...
+  private final String toEmail;
+
+  /** The notification recipient */
+  private final String recipientName;
+
+  private final Gmail service;
+
+  /**
+   * @param recipientName Name of the recipient of the user
+   * @param toEmail The email address that this email is going to be sent to.
```
kmh287, line 73: please add the new gmail param to the docs as well

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /** Getter function that returns the string representing the email recipients name . */
+  public String getRecipientName() {
+    return recipientName;
+  }
+
+  public Gmail getService() {
+    return service;
+  }
```
kmh287, line 181: please remove these getters unless they're needed by calling production code

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,159 @@
...
+  @Test
+  public void constructorShouldSetNameAndEmailTo() {
+    assertEquals("John", emailNotifier.getRecipientName());
+    assertEquals("jdoe@gmail.com", emailNotifier.getToEmail());
+  }
+
+  @Test
+  public void testMessageHasCorrectApplicationName() throws MessagingException, IOException {
+    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
+    emailNotifier.getService().users().messages().send("anyString", testMessage);
```
kmh287, line 98: If the test accesses the gmail service directly and sends the message then we're missing out on testing the code within the EmailNotifier class.

Please make sure that this test and all the others are calling notifyUser.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Global instance of the scopes required by this quickstart. If modifying these scopes, delete
+   * your previously saved tokens/ folder.
+   */
+  private static final List<String> SCOPES = ImmutableList.of(GmailScopes.MAIL_GOOGLE_COM);
+
+  private static final String CREDENTIALS_FILE_PATH = "backend/credentials.json";
+
+  /** The email to be notified */
+  private final String toEmail;
```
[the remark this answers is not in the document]
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: I am not using them within the email notifier Itself but I am currently using all three of these getters within the tests .

kmh287: Right, since they're not used by any of the non-test code we want to avoid needlessly exposing the internals of our class. Since we have a way to get the toEmail and recipientName from the sent email on the Gmail mocks and the gmail service is instantiated in the test.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: i believe I have addressed these issues via my understanding of it !

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,178 @@
...
+    MimeMessage mimemessage = createTestMessage();
+    testMessage = createMessageFromMime(mimemessage);
+    // Test Instance of email notifier class.
+    emailNotifier = new EmailNotifier("John", "jdoe@gmail.com ", gmail);
+    // Emulating real behavior due to when method, will return list of users.
+    when(gmail.users()).thenReturn(users);
+    // Emulating real behaviour due to when method, will return list of messages
+    when(users.messages()).thenReturn(messages);
+    //    Emulating sending messages, when sent will return send mock.
+    when(messages.send(anyString(), any(Message.class))).thenReturn(send);
```
kmh287, line 57: nit: We can use any(), any() instead of explicitly specifying the types unless the send method has conflicting overloads. 

These methods are an artifact of support for pre-Java 7 in which the matchers had to explicitly specify the type. In java 7 and later these don't have any effect. For example, `any(Message.class)` doesn't actually restrict matches to just `Message` and its subclasses.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -0,0 +1,161 @@
...
+   * Global instance of the scopes required by this quickstart. If modifying these scopes, delete
+   * your previously saved tokens/ folder.
+   */
+  private static final List<String> SCOPES = ImmutableList.of(GmailScopes.MAIL_GOOGLE_COM);
+
+  private static final String CREDENTIALS_FILE_PATH = "backend/credentials.json";
+  /** The email to be notified */
+  public final String recipientEmail;
+  /** The notification recipient */
+  public final String recipientName;
```
kmh287, line 65: private

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,178 @@
...
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+import org.apache.commons.codec.binary.Base64;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+
+public class EmailNotifierTest {
```
kmh287, line 32: Please add the @RunWith annotation

@RunWith(JUnit4.class)

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,178 @@
...
+  private EmailNotifier emailNotifier;
+  @Mock private Gmail gmail;
+  @Mock private Gmail.Users users;
+  @Mock private Gmail.Users.Messages messages;
+  @Mock private Gmail.Users.Messages.Send send;
```
kmh287, line 37: Cool, I'm glad you found the @Mock annotation! This is useful not only for specifying that certain fields are really mocks but can also help set them up. If you add a call to `Mockito.initMocks(this)` in the setUp method, the mocks for these four fields will be automatically created and assigned. 

Up to you if you want to use @Mock that way or remove the annotations and keep what's currently in setUp.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,178 @@
...
+    Properties properties = System.getProperties();
+    Session session = Session.getDefaultInstance(properties);
+    byte[] bytes = msg.decodeRaw();
+    InputStream targetStream = new ByteArrayInputStream(bytes);
+    return new MimeMessage(session, targetStream);
+  }
+
+  @Test
+  public void constructorShouldSetNameAndEmail() {
+    assertEquals("John", emailNotifier.recipientName);
```
kmh287, line 100: I think I commented earlier in the PR but I can't find it so apologies if I missed a response. Please convert all assertions to [Truth](www.truth.dev)

kmh287, line 100: Apologies if I asked this earlier and missed a reply. Please convert the assertions in this test to Truth. https://truth.dev/

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Done

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: IN the case of no getters , these field being private and me not mocking the test that I am writing I am unsure of the constructor test that I have created and whether I can still create it

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: I believe I fixed

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    // Test Instance of email notifier class.
+    emailNotifier = new EmailNotifier("John", "jdoe@gmail.com ", gmail);
+    // Emulating real behavior due to when method, will return list of users.
+    when(gmail.users()).thenReturn(users);
+    // Emulating real behaviour due to when method, will return list of messages
+    when(users.messages()).thenReturn(messages);
+    //    Emulating sending messages, when sent will return send mock.
+    when(messages.send(any(), any())).thenReturn(send);
+    //    When send mock is executed will return test message.
+    when(send.execute()).thenReturn(testMessage);
```
kmh287, line 60: Since the tests now use the message constructed in EmailNotifier, can this line be removed?

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    Session session = Session.getDefaultInstance(properties);
+    byte[] bytes = msg.decodeRaw();
+    InputStream targetStream = new ByteArrayInputStream(bytes);
+    return new MimeMessage(session, targetStream);
+  }
+
+  @Test
+  public void testMessageHasCorrectApplicationName()
+      throws MessagingException, IOException, GeneralSecurityException {
+    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
```
kmh287, line 102: nit: Please split all tests into distinct arrange, act, and assert blocks that are each separated by a newline. 

```
  @Test
  public void testMessageHasCorrectApplicationName()
      throws MessagingException, IOException, GeneralSecurityException {
    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);

    emailNotifier.notifyUser();

    verify(messages).send(any(), argument.capture());
    String applicationName = convertToMimeMessage(argument.getValue()).getFrom()[0].toString();
    Truth.assertWithMessage("These two strings should be the same")
        .that("Adlib-Step@gmail.com")
        .isEqualTo(applicationName);
  }
```

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    emailNotifier.notifyUser();
+    verify(messages).send(any(), argument.capture());
+    String applicationName = convertToMimeMessage(argument.getValue()).getFrom()[0].toString();
+    Truth.assertWithMessage("These two strings should be the same")
```
kmh287, line 106: nit: the assertWithMessage method should be used to add additional context to the error message. Truth will already report an error that the two Strings are not equal.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    verify(messages).send(any(), argument.capture());
+    Address[] allRecipients = convertToMimeMessage(argument.getValue()).getAllRecipients();
+    Address[] correctRecipients =
+        new Address[] {
+          new InternetAddress("jdoe@gmail.com "),
+        };
+    Truth.assertWithMessage("These Two arrays should be the same ")
+        .that(correctRecipients)
+        .isEqualTo(allRecipients);
+    Truth.assertThat(allRecipients[0]).isEqualTo(correctRecipients[0]);
```
kmh287, line 163: assertThat(allRecipients).containsExactly(new InternetAddress("jdoe@gmail.com"))

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    gmail = mock(Gmail.class);
+    // Mock to return list of users
+    users = mock(Gmail.Users.class);
+    // Mock to return list of messages.
+    messages = mock(Gmail.Users.Messages.class);
+    send = mock(Gmail.Users.Messages.Send.class);
+    MimeMessage mimemessage = createTestMessage();
+    testMessage = createMessageFromMime(mimemessage);
+    // Test Instance of email notifier class.
+    emailNotifier = new EmailNotifier("John", "jdoe@gmail.com ", gmail);
```
kmh287, line 52: Please extract these Strings to constants that can be referenced elsewhere in the test. For example "jdoe@gmail.com" is used here and below. If extracted to a private static final String constant, it would be easy to change the email address in one place without searching and replacing everywhere within the test.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+  @Test
+  public void testMessageHasCorrectApplicationName()
+      throws MessagingException, IOException, GeneralSecurityException {
+    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
+    emailNotifier.notifyUser();
+    verify(messages).send(any(), argument.capture());
+    String applicationName = convertToMimeMessage(argument.getValue()).getFrom()[0].toString();
+    Truth.assertWithMessage("These two strings should be the same")
+        .that("Adlib-Step@gmail.com")
+        .isEqualTo(applicationName);
```
kmh287, line 108: looks like expected and actual are flipped here and below

`assertThat(actual).isEqualTo(expected)`

Since "Adlib-Step@gmail.com" is the expected value, it should go second.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,177 @@
...
+  public void testMessageShouldHaveIncorrectRecipients()
+      throws MessagingException, IOException, GeneralSecurityException {
+    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
+    emailNotifier.notifyUser();
+    verify(messages).send(any(), argument.capture());
+    Address[] allRecipients = convertToMimeMessage(argument.getValue()).getAllRecipients();
+    Truth.assertWithMessage("These Two arrays should be the same ")
+        .that(new InternetAddress[] {new InternetAddress()})
+        .isNotEqualTo(allRecipients);
+  }
```
kmh287, line 176: Can you elaborate on the scenario this test is testing? The test above checks that in this scenario allRecipients will not be empty right?

TcgVanguardTroll: The test above this one asserted that the given Attendees were correct, given that the current implementation is only sending an email to one person thus would need two function calls in the future this test asserts that given that the attendees cannot be empty given the constraints of the functions being called I am comparing the mock with behavior that is never expected to be present

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Deleted line 59-60

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Fixed Function Calls

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Swapped tests expected and actual

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Does that that clear some of your confusion ?

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Extracted

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
kmh287: Attendees == recipients?

But is it possible for this test to fail when the other test doesn't fail? The one above has the same setup and action but asserts that the recipients has a single entry. That would also seem to cover this test method's assertion that the recipients are not empty.

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,168 @@
...
+    testEmail = "jdoe@gmail.com";
+    testName = "John";
+    testString = "Text Text";
+    gmail = mock(Gmail.class);
+    // Mock to return list of users
+    users = mock(Gmail.Users.class);
+    // Mock to return list of messages.
+    messages = mock(Gmail.Users.Messages.class);
+    send = mock(Gmail.Users.Messages.Send.class);
+    MimeMessage mimemessage = createTestMessage();
```
kmh287, line 52: Are the variable and the createTestMessage method still needed?

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import com.google.api.services.gmail.Gmail;
+import com.google.api.services.gmail.model.Message;
+import com.google.common.truth.Truth;
```
kmh287, line 11: optional: this can be imported statically (to remove the Truth. prefix to all the method calls) just as the mockito methods are above

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Refactored

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Imported statically

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Yes I meant recipients instead of attendees. After your explanation i can see where the test case is redundant. Acknowledged

backend/src/test/java/com/google/sps/EmailNotifierTest.java
kmh287: ping

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+  @Test
+  public void testMessageShouldHaveIncorrectBodyText()
+      throws MessagingException, IOException, GeneralSecurityException {
+    ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
+
+    emailNotifier.notifyUser();
+
+    verify(messages).send(eq("me"), argument.capture());
+
```
kmh287, line 117: tiny nit: remove the extra newline here

TcgVanguardTroll: Removed

tcwanggoogle:
Thank you Jordan for getting through this PR and thanks Kevin for providing the instructions. Just one more comment here to Jordan. Please wait for all the checks to pass before merging the branch for the next time. Otherwise we are not sure if a check fails in the next PR is due to which PR.

PR #19, opened 2020-07-08
Merge working-jan-dev/jan-dev changes into master

jzwang8:
All changes have already been reviewed and approved in the latest closed PR, but I need approval from everyone to merge with master!


backend/pom.xml
```diff
@@ -51,9 +57,9 @@
     </dependency>
     
     <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <version>3.3.3</version>
```
kmh287, line 56: Are there no tests using mockito at the moment? If so it's ok to remove this for now.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+
+/** Class used to find a match in a list of Participants with the most recently added Participant */
+public final class FindMatchQuery {
+
+  /** Maximum difference in duration to be compatible */
+  private static int MAX_DURATION_DIFF = 15;
+  /** extra padding time to ensure large enough meeting time block */
+  private static int PADDING_TIME = 15;
+  /** Reference clock for "current" time */
```
kmh287, line 30: tiny nit: Now that it's a clock and not a ZoneDateTime, returning the "current time" is intrinsic to the type (clocks tell you what time it is right now) so this javadoc can be shortened or removed.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,85 @@
...
+  private Clock clock;
+
+  /** Set "current" date and time to calculate whether a match is possible in time */
```
kmh287, line 33: same comment about the javadoc here

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -0,0 +1,85 @@
...
+    ZonedDateTime dateTime = ZonedDateTime.now(clock);
+
+    // Compare new participant preferences with others in list to find match
+    for (Participant currParticipant : participants) {
+
+      // Check if participants are looking for similar meeting duration
+      int newDuration = newParticipant.getDuration();
+      int currDuration = currParticipant.getDuration();
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
+      int duration = Math.min(newDuration, currDuration);
```
kmh287, line 53: tiny nit: this can go below the if since it's not used if compatibleDuration is false

backend/src/main/java/com/google/sps/data/TimeHelper.java
```diff
@@ -0,0 +1,29 @@
...
+ * timestamp
+ */
+public final class TimeHelper {
+
+  private static ZonedDateTime dateTime;
+
+  /** Constructor */
+  public TimeHelper() {
+    this(ZonedDateTime.now());
+  }
```
kmh287, line 16: Sorry I missed this earlier. I'd recommend eliminating this constructor and only using the one below. When classes construct their own dependencies (e.g. this class constructing its ZoneDateTime internally here in the constructor) they become harder to test. Using "dependency injection" as you have in the constructor below allows a class' dependencies to be provided from the outside which in turn makes them easier to test.

 Since you don't want to test one code path (the constructor that accepts a ZoneDateTime) while using another in prod, I would recommend simplifying down to one.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    Instant endTimeAvailableInstant =
+        Instant.ofEpochMilli(
+            Long.parseLong(
+                request.getParameter("endTimeAvailable"))); // TODO: figure out input format
+    LocalDateTime endTimeAvailableLocal =
+        endTimeAvailableInstant.atZone(ZoneId.of("UTC")).toLocalDateTime();
+    ZonedDateTime endTimeAvailable = endTimeAvailableLocal.atZone(zoneId);
+
+    int duration = convertToPositiveInt(request.getParameter("duration"));
+    if (duration == -1) {
```
kmh287, line 83: Functionally the same, but <= 0 seems clearer. WDYT?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    if (email == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
+      return;
+    }
+    String username = email.split("@")[0];
+
+    String timezone = request.getParameter("timezone");
+    ZoneId zoneId = ZoneId.of(timezone); // TODO: convert input timezone to valid ZoneId
+    ZonedDateTime startTimeAvailable =
+        LocalDateTime.now().atZone(zoneId); // TODO: set to future time if not available now
```
kmh287, line 73: https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#now-java.time.ZoneId-

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    // TODO: Check input values are filled (before allowing to submit)
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    if (email == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
+      return;
+    }
+    String username = email.split("@")[0];
+
+    String timezone = request.getParameter("timezone");
```
kmh287, line 70: if I enter the system from EST and someone else enters from PST, how are those times "normalized" in the system? If I'm free for an hour from 3-4 and someone from the west coast is free for an hour from 12-1 then we should be a match even if the times / timestamps are different.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    ZoneId zoneId = ZoneId.of(timezone); // TODO: convert input timezone to valid ZoneId
+    ZonedDateTime startTimeAvailable =
+        LocalDateTime.now().atZone(zoneId); // TODO: set to future time if not available now
+    Instant endTimeAvailableInstant =
+        Instant.ofEpochMilli(
+            Long.parseLong(
+                request.getParameter("endTimeAvailable"))); // TODO: figure out input format
```
kmh287, line 77: needs to be zoned

backend/pom.xml
jzwang8: I did remove it before. Is it showing up on your end? On my end, I see those lines as removed.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: For endTimeAvailable (ZonedDateTime), we take in their timezone to calculate the time they're available to on that day (depends on what frontend sends back for input). startTimeAvailable (ZonedDateTime) is just the current time but in their timezone that they explicitly specified. The only calculation using times to see if they're a match is using ZDT1.isBefore(ZDT2)  which takes into account timezone differences.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: For endTimeAvailable, I first create an instant assuming endTimeAvailable is given in milliseconds (depends on the input format frontend sends), create a LocalDateTime using milliseconds, and create a ZonedDateTime using the LocalDateTime and ZoneId.

backend/pom.xml
kmh287: I seem them as removed by this PR. I'm saying it's ok to do that now but it will almost certainly be added back in the near-future as more backend tests are added.

jzwang8: Oh, I get it. I don't have any mockito tests right now so I removed the dependency.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    ZoneId zoneId = ZoneId.of(timezone); // TODO: convert input timezone to valid ZoneId
+    ZonedDateTime startTimeAvailable =
+        ZonedDateTime.now(zoneId); // TODO: set to future time if not available now
+    Instant endTimeAvailableInstant =
+        Instant.ofEpochMilli(
+            Long.parseLong(
+                request.getParameter("endTimeAvailable"))); // TODO: figure out input format
+    LocalDateTime endTimeAvailableLocal =
+        endTimeAvailableInstant.atZone(ZoneId.of("UTC")).toLocalDateTime();
+    ZonedDateTime endTimeAvailable = endTimeAvailableLocal.atZone(zoneId);
```
kmh287, line 80: Would this work?

`ZoneDateTime endTimeAvailable = endTimeAvailableInstant.atZone(zoneId)`

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
kmh287: Ah, got it. Thanks

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
kmh287: understood, thanks

backend/src/main/java/com/google/sps/data/TimeHelper.java
```diff
@@ -0,0 +1,24 @@
...
+
+  private static ZonedDateTime dateTime;
+
+  /** Constructor */
+  public TimeHelper(ZonedDateTime dateTime) {
+    this.dateTime = dateTime;
+  }
+
+  /** Return today's date with time of hour:minute */
+  public static ZonedDateTime getNewTimeToday(int hour, int minute) {
```
kmh287, line 19: Having an instance constructor set a static field is error prone

```
TimeHelper th1 = TimeHelper(/* December 4, 2019 */)
TimeHelper th2 = TimeHelper(/* June 27, 2053 */)
th1.getNewTimeToday(12, 25)
```

What does that evaluate to? I would expect 2019-12-04 12:25 but it would actually be 2053-06-27 12:25.

I recommend making both `dateTime` and `getNewTimeToday` non-static.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,184 @@
...
+  private static final String PERSON_B = "Person B";
+  private static final String PERSON_C = "Person C";
+  private static final String PERSON_D = "Person D";
+
+  // Default parameters unused in query
+  private static final long ID_DEFAULT = 0;
+  private static final long TIMESTAMP_DEFAULT = 0;
+
+  // Some times available until on 1/1/2020
+  private static final ZonedDateTime TIME_0200PM = TimeHelper.getNewTimeToday(14, 0);
```
kmh287, line 45: since TimeHelper's constructor isn't called in this test, shouldn't this cause a null pointer exception? That's another reason to make getNewTimeToday an instance method.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,184 @@
...
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class FindMatchQueryTest {
```
kmh287, line 32: Please add test cases to this test that verify that the algorithm will work for participants in different timezones

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
kmh287: Understood. Let's make sure that's tested.

dhruvi618, line 70: Currently, the frontend is not necessarily asking for user input on the timezone and its being detected by JS Date class (which can be converted to the Java ZonedDateTime class you're using). I was wondering if this would parameter request for still be needed in that case?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -0,0 +1,65 @@
...
+      ZonedDateTime startTimeAvailable,
+      ZonedDateTime endTimeAvailable,
+      int duration,
+      long timestamp) {
+    this.id = id;
+    this.username = username;
+
+    Preconditions.checkArgument(
+        startTimeAvailable.isBefore(endTimeAvailable),
+        "Start available time must be before end available time.");
```
dhruvi618, line 35: This is really cool, I didn't know you could do this before your PR

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Nope it wouldn't be necessary, but I wasn't sure what format the backend would get when I wrote this code. We can work together after to make sure all input formats match

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.simple.JSONObject;
+
+/** Servlet that returns some example content. */
+@WebServlet("/search-match")
```
dhruvi618, line 37: Based on the project structure changes I made yesterday, in order to correctly route requests between front-end and back-end, the servlet endpoint will need to be renamed with the prefix /api/v1. For example, "search-match" changes to be "/api/v1/search-match"

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -0,0 +1,176 @@
...
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns some example content. */
+@WebServlet("/add-participant")
```
dhruvi618, line 44: Same comment here about adding the prefix `/api/v1` to the servlet

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: Added test cases!

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+    Match match = query.findMatch(Arrays.asList(participantA), participantB);
+
+    assertThat(match.getFirstParticipant().getUsername()).isEqualTo(PERSON_B);
+    assertThat(match.getSecondParticipant().getUsername()).isEqualTo(PERSON_A);
+    assertThat(match.getDuration()).isEqualTo(DURATION_15_MINUTES);
+  }
+
+  /** Return today's date with time of hour:minute */
+  public ZonedDateTime getNewTimeToday(ZonedDateTime dateTime, int hour, int minute) {
```
kmh287, line 115: nit: private static. That should also allow the initialization to move to the declaration, removing the corresponding setup from the setUp method

private static final ZonedDateTime TIME_1450ET = getNewTimeToday(currentDateTimeET, 14, 50);

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+  private static final String PERSON_B = "Person B";
+  private static final String PERSON_C = "Person C";
+  private static final String PERSON_D = "Person D";
+
+  // Default parameters unused in query
+  private static final long ID_DEFAULT = 0;
+  private static final long TIMESTAMP_DEFAULT = 0;
+
+  // Some times available until on 1/1/2020
+  private ZonedDateTime TIME_1400ET;
```
kmh287, line 44: nit: these should be static final constants as the ones above. See my comment below on how to do that.

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+    assertThat(match.getFirstParticipant().getUsername()).isEqualTo(PERSON_B);
+    assertThat(match.getSecondParticipant().getUsername()).isEqualTo(PERSON_A);
+    assertThat(match.getDuration()).isEqualTo(DURATION_30_MINUTES);
+  }
+
+  @Test
+  public void compatibleWithoutConsideringTimezone() {
+    // Two participants, A in ET timezone and B in PT timezone, compatible if not considering
+    // timezones, compatible if considering their different timezones
```
kmh287, line 264: They are compatible *only* because their timezones have been considered. Otherwise 1400 - 1500 and 1100 - 1200 wouldn't overlap.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Ok sounds good!

PR #14, opened 2020-06-30
JS linter: Add initial support for JS linter

tcwanggoogle:
Only lint JS files under frontend/

PR #20, opened 2020-07-08
Fix lint errors in all JS files 

dhruvi618:


tcwanggoogle:
Can you merge this PR into js-linter (https://github.com/googleinterns/Ad-lib/pull/14), or the other way? In this case, we will have the linter in the PR, so we know it passes or not.

PR #21, opened 2020-07-09
Update user interface and retrieve user inputs on form submit

dhruvi618:
UI Design: https://screenshot.googleplex.com/u6xcWJ75gVP

Next Steps: Add JSON object and execute POST to backend and add tests


frontend/src/components/ProductAreaDropdown.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import InputLabel from '@material-ui/core/InputLabel';
+import MenuItem from '@material-ui/core/MenuItem';
+import FormControl from '@material-ui/core/FormControl';
+import ListItemText from '@material-ui/core/ListItemText';
+import Select from '@material-ui/core/Select';
+
+// TO-DO: Add more product area options to this array
```
kmh287, line 9: an internal search for "list of PAs" should give the full list

kmh287:
Just like how ERG could be a proxy for certain personal information, years at google could also be a proxy for someone's age. How important is "years at google" data point to the planned matching algorithm? 

 @jzwang8

frontend/src/components/MenuBar.js
```diff
@@ -6,6 +6,9 @@ import Typography from '@material-ui/core/Typography';
 import IconButton from '@material-ui/core/IconButton';
 import HelpIcon from '@material-ui/icons/Help';
 
+/**
+ * Establishes style to use on rendering menu bar component
+ */
```
jzwang8, line 11: I think one line javadocs should just be /** comment */

jzwang8:
> Just like how ERG could be a proxy for certain personal information, years at google could also be a proxy for someone's age. How important is "years at google" data point to the planned matching algorithm?
> 
> @jzwang8

If we remove both ERG and Years @, then we only have 2 matching fields left, so we're planning on thinking of new areas to use for matching or maybe we'll decide to completely remove those fields and just do all random?

kmh287:
For the prototype random should be fine. For MVP I remember we discussed allowing users to indicate interests. That could be a source for a lot of data points without asking for potentially sensitive personal information.

frontend/.gcloudignore
```diff
@@ -0,0 +1,17 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
```
tcwanggoogle, line 1: I think this file should already be in master? Can you try to rebase the branch to see if these diff are removed in the PR?

frontend/src/components/ProductAreaDropdown.js
```diff
@@ -0,0 +1,48 @@
...
+const productAreas = [
+  'Play',
+  'Search',
+  'Ads',
+  'Youtube',
+  'Maps',
+];
```
tcwanggoogle, line 16: Add 'Core' (to promote my own PA...)

frontend/src/components/YearsDropdown.js
```diff
@@ -0,0 +1,47 @@
...
+YearDropdown.propTypes = {
+  onChange: PropTypes.func,
+};
+
+/**
+ * Create year dropdown component using yearRanges array
+ * @return {YearDropdown} YearDropdown component
+ * @param {Object} props
+ */
+export default function YearDropdown(props) {
```
tcwanggoogle, line 27: Any ways we can some how combine these *Dropdown() into a single function buy taking different label and inputs? This should make the code easier to extend with more fields, as I see these dropdowns are very similar

frontend/src/components/ProductAreaDropdown.js
dhruvi618: Added! Thanks for the resource!

frontend/src/components/MenuBar.js
dhruvi618: Fixed in the MenuBar.js file and Form.js file, thank you!

frontend/.gcloudignore
dhruvi618: I'll remove this file from this PR, but also I think we can take it out of master as well. I now realize that it's added every time the webapp is deployed

tcwanggoogle: I see. So you can add `frontend/.gcloudignore` or add `**/.gcloudignore` to `.gitignore` so future PR won't include it

frontend/src/components/YearsDropdown.js
dhruvi618: I was initially trying to do this exact thing by taking different labels and inputs within, but problems arise when trying to get the state of each component on submit and the `id` of each dropdown would be the same

frontend/src/components/ProductAreaDropdown.js
dhruvi618: Added and updated to include all PAs!

frontend/src/components/YearsDropdown.js
tcwanggoogle: Ok! (Since I am not familiar with React or JS, I don't know how to fix it. So...) I think it's OK to leave as is

.gcloudignore
```diff
@@ -1,17 +0,0 @@
-# This file specifies files that are *not* uploaded to Google Cloud Platform
```
kmh287, line 1: Revert deletion of this file?

frontend/src/components/ProductAreaDropdown.js
kmh287: woo P&E!

.gcloudignore
dhruvi618: I realized this file is added every time the project is deployed to gcloud, so I was thinking we could exclude it from the repo?

kmh287: Seems prudent to keep it around in the repo then right? Similar to how .gitignore is included.

dhruvi618: Ok, I'll revert the deletion for it. Just wondering if its better to include the file in the same directory it is automatically generated in (`/frontend`) or the main project directory?

kmh287: Do you know why it generates in the frontend directory? Is it because that's the directory you have to deploy from?

dhruvi618: Yeah, that's where I deploy from

kmh287: If you deployed from the root would the file be generated there or always under /frontend? If so I'd say go ahead and delete this in favor of one under /frontend

dhruvi618: Ok, got it. Since it only allows you to deploy from /frontend, I will favor that file and keep this deleted

PR #22, opened 2020-07-10
Create Match and Participant Datastore classes

jzwang8:
Helpful definitions:
- Key = unique Key used to identify specific entities in the datastore
- ID = unique number (type long) that corresponds to a Key 
- Kind = type of entity (either Match or Participant)
- Property = fields for an entity (duration, endTimeAvailable, firstParticipantKey, etc.)

Changes from my last PR:
- Created MatchDatastore and ParticipantDatastore to separate direct Datastore access from servlets and matching
- Use unique key ID for each Match and each Participant for identification purposes
- 1 Participant = 1 person, even if they participate multiple different times (used to have multiple participants corresponding to 1 person with same username)
- Every match is stored in the datastore, never deleted. All participants are stored in the datastore, never deleted (If they're not currently unmatched, then their availability fields are nulled out. When we incorporate additional fields for the MVP, these can be nulled out or saved depending on user preference.)
- A Participant has a reference to their current match. If no match has been found yet, reference is null. If match has been found AND not "returned" (will eventually be changed to "notified" once we incorporate notifications), reference is the match Key in the datastore.
- FindMatchQueryTest tests datastore functionality now since FindMatchQuery uses datastore.

Questions for future development:
- Any tests I can add?
- I have a lot of methods in ParticipantDatastore. Do you see any way I can make them clearer, reduce any, etc.?
- I had trouble defining what should be done in FindMatchQuery vs in AddParticipantServlet. Any suggestions or feedback?


backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -36,23 +37,29 @@ public FindMatchQuery(Clock clock) {
   }
 
   /**
-   * Find match of new participant or add to list of participants, return whether or not match was
-   * found right after being added
+   * Find and return match of new participant with unmatched participants by comparing duration and
+   * availibility or return null if no match yet
    */
-  public Match findMatch(List<Participant> participants, Participant newParticipant) {
+  public Match findMatch(ParticipantDatastore participantDatastore, Participant newParticipant) {
```
kmh287, line 43: WDYT about making ParticipantDatastore a constructor parameter and using it as a field instead of a method arg?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -36,23 +37,29 @@ public FindMatchQuery(Clock clock) {
...
+    // Get list of unmatched participants in datastore
+    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
+
+    // Add newParticipant to datastore
+    participantDatastore.addParticipant(newParticipant);
```
kmh287, line 48: This seems like an unexpected side effect of this method. Could it be pushed up to the callers? Should a participant who is matched by this algorithm be inserted into the datastore at all?

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -1,34 +1,27 @@
...
+  /** Match datastore key ID */
   private final long id;
-  /** First participant */
-  private final Participant firstParticipant;
-  /** Second participant */
-  private final Participant secondParticipant;
+  /** First participant datastore key */
+  private final Key firstParticipantKey;
+  /** Second participant datastore key */
+  private final Key secondParticipantKey;
```
kmh287, line 13: This change breaks the abstraction of the new *store classes. Ideally those classes entirely encapsulate interactions with appengine.api.datastore and so passing datastore keys around in these model classes ties them to a specific datastore implementation. With perfect encapsulation, one could refactor the *store classes within this PR to use something other than appengine.api.datastore for storage without affecting any other code.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.sps.data.Match;
+
+public final class MatchDatastore {
+
+  /** Datastore */
+  private DatastoreService datastore;
```
kmh287, line 13: nit: Instance fields should always appear after static fields

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,70 @@
...
+    // Return matchKey
+    return matchEntity.getKey();
+  }
+
+  /** Return match based on match datastore key */
+  public Match getMatchFromKey(Key matchKey) {
+    // No match
+    if (matchKey == null) {
+      return null;
+    }
```
kmh287, line 48: Good opportunity here to start introducing nullability annotations to our java code.
https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305/3.0.2

You can mark fields, parameters, and method return types as Nullable or NotNull using annotations. The compiler doesn't enforce these, but linters and other analysis tools can add warnings to code for example that passes null as an argument to a method where the parameter is marked as @NonNull. 

The annotations also help define the contract for your method. I would argue here that requesting something from a datastore using a null key is something the caller should guard against, not the datastore. I recommend adding a @NonNull annotation to the parameter and a @Nullable annotation to the method.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Done. Also changed in tests and still passes.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Moved it to AddParticipantServlet. I put it into the match method for the tests but I found an easy way to test still.

backend/src/main/java/com/google/sps/data/Match.java
jzwang8: The problem I ran into is that a Participant is an invalid property type (only types like int, long, Key, etc. can be stored). An alternative is to have a Match consist of actual Participants, but only store the Participant Keys in the datastore. Then MatchDatastore would depend on ParticipantDatastore to switch between Match objects and Match entities. Do you think that would be better?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: Added @Nullable annotations to methods that return null. For the part you pointed out above, I realized I did check if it was null in the caller, so I removed the null check.

backend/src/main/java/com/google/sps/data/Match.java
kmh287: I like the idea of pulling Participant out of this class. Match and Participant are two separate models that are stored separately. While Match has to reference the matched participants, its model doesn't necessarily need to contain the full Participant object. 

This is where you would introduce String keys or long IDs for your Participants that aren't tied to the storage implementation. User emails make sense as keys to me.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: Also, I changed it so that all participants, once inserted into the datastore, stay forever. Their inputs are just nulled out if a match has been found or their time is up.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,55 @@
...
+  public Match findMatch(Participant newParticipant) {
+    // Get list of unmatched participants in datastore
+    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
+
+    // Add newParticipant to datastore
+    participantDatastore.addParticipant(newParticipant);
+
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
-    System.out.println(dateTime);
```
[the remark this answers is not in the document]
kmh287: +1. I'd encourage replacing println statements with calls to a logger.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,55 @@
...
 
+    // Compare new participant preferences with other participants with similar duration to find
+    // match
+    for (Participant currParticipant : unmatchedParticipants) {
       // Check if participants are looking for similar meeting duration
       int newDuration = newParticipant.getDuration();
       int currDuration = currParticipant.getDuration();
       boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
 
       if (!compatibleDuration) {
```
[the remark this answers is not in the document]
kmh287: either way is fine by the style guide. I like to pull out variables like this especially when doing so helps the if condition fit on a single line

backend/src/main/java/com/google/sps/data/Match.java
jzwang8: I changed Key references to be String usernames for each Match, and for each participant, the match reference is now just the long id.

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
jzwang8: Got it. Thanks, will add in next PR.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,54 @@
...
-  public FindMatchQuery(Clock clock) {
+  public FindMatchQuery(Clock clock, ParticipantDatastore participantDatastore) {
     this.clock = clock;
+    this.participantDatastore = participantDatastore;
   }
 
   /**
-   * Find match of new participant or add to list of participants, return whether or not match was
-   * found right after being added
+   * Find and return match of new participant with unmatched participants by comparing duration and
```
kmh287, line 44: @return "Match of new participant....

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,54 @@
...
 
   /** Maximum difference in duration to be compatible */
   private static int MAX_DURATION_DIFF = 15;
-  /** extra padding time to ensure large enough meeting time block */
+  /** Extra padding time to ensure large enough meeting time block */
   private static int PADDING_TIME = 15;
   /** Reference clock */
   private Clock clock;
+  /** Datastore of Participants */
+  private ParticipantDatastore participantDatastore;
```
kmh287, line 35: these can both be marked as final

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,54 @@
...
 
-/** Class used to find a match in a list of Participants with the most recently added Participant */
+/** Class used to find a match for new Participant with unmatched Participants in datastore */
 public final class FindMatchQuery {
 
   /** Maximum difference in duration to be compatible */
   private static int MAX_DURATION_DIFF = 15;
-  /** extra padding time to ensure large enough meeting time block */
+  /** Extra padding time to ensure large enough meeting time block */
   private static int PADDING_TIME = 15;
```
kmh287, line 31: final

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -3,46 +3,41 @@
...
 
   /** Initialize constructor fields */
   public Participant(
-      long id,
       String username,
       ZonedDateTime startTimeAvailable,
       ZonedDateTime endTimeAvailable,
       int duration,
+      long currentMatchId,
       long timestamp) {
```
kmh287, line 29: Once a Participant has a match, should they be removed from the system? If so, why add a field for currentMatchId?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,94 @@
...
+
+  // Datastore Key/Property constants
+  private static final String KIND_MATCH = "Match";
+  private static final String PROPERTY_FIRSTPARTICIPANTUSERNAME = "firstParticipantUsername";
+  private static final String PROPERTY_SECONDPARTICIPANTUSERNAME = "secondParticipantUsername";
+  private static final String PROPERTY_DURATION = "duration";
+  private static final String PROPERTY_TIMESTAMP = "timestamp";
+
+  /** Datastore */
+  private DatastoreService datastore;
```
kmh287, line 23: final

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,94 @@
...
+    Key matchKey = KeyFactory.createKey(KIND_MATCH, matchId);
+    try {
+      return getMatchFromEntity(datastore.get(matchKey));
+    } catch (EntityNotFoundException e) {
+      return null;
+    }
+  }
+
+  /** Return Match from Entity */
+  public Match getMatchFromEntity(Entity matchEntity) {
```
kmh287, line 59: private? Making this public exposes the use of appengine datastore through the Entity parameter

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,94 @@
...
+    String secondParticipantUsername =
+        (String) matchEntity.getProperty(PROPERTY_SECONDPARTICIPANTUSERNAME);
+    int duration = ((Long) matchEntity.getProperty(PROPERTY_DURATION)).intValue();
+    long timestamp = (long) matchEntity.getProperty(PROPERTY_TIMESTAMP);
+
+    return new Match(id, firstParticipantUsername, secondParticipantUsername, duration, timestamp);
+  }
+
+  /** Remove Match from datastore */
+  public void removeMatch(Match match) {
```
kmh287, line 72: Looks like this only needs the ID

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Once a Participant has a match, their availability fields and currentMatchId are nulled out, but they still remain in the datastore. We decided not to remove participants from the datastore since participants have the option to save their preferences. If they choose to save their preferences, their availability and matchId will be nulled out each time their match is successfully returned to them, but their other personal preferences (role, PA, interests) will be saved. If they choose NOT to save their preferences, we still keep them in the datastore as a participant but null out ALL fields. (Right now, the only fields are time availability and matchId).

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,150 @@
...
+      return datastore.get(participantKey);
+    } catch (EntityNotFoundException e) {
+      // TODO: what error to send
+      return null;
+    }
+  }
+
+  /** Update Participant with new matchId, null out availability to show currently matched */
+  public void updateNewMatch(String username, long matchId) {
+    Entity participantEntity = getEntityFromUsername(username);
```
dhruvi618, line 73: Since `getEntityFromUsername` can return null on an EntityNotFoundException, what do you think about adding a null check here?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,150 @@
...
+    participantEntity.setProperty(PROPERTY_DURATION, 0);
+
+    // Overwrite existing entity in datastore
+    datastore.put(participantEntity);
+  }
+
+  /** Return Participant from username */
+  @Nullable
+  public Participant getParticipantFromUsername(String username) {
+    return getParticipantFromEntity(getEntityFromUsername(username));
```
dhruvi618, line 89: Similar comment here about potentially adding a null check. In the case where this does return null, will the algorithm still behave expectedly? Maybe adding a test for this case might be helpful, what do you think?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,94 @@
...
+    long timestamp = (long) matchEntity.getProperty(PROPERTY_TIMESTAMP);
+
+    return new Match(id, firstParticipantUsername, secondParticipantUsername, duration, timestamp);
+  }
+
+  /** Remove Match from datastore */
+  public void removeMatch(long matchId) {
+    Key matchKey = KeyFactory.createKey(KIND_MATCH, matchId);
+    datastore.delete(matchKey);
+  }
```
dhruvi618, line 75: When will this `removeMatch` method be called (does this happen when both participants have been notified)? If so, would we also need to null out the matchID fields of both participants in the participant datastore?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /** Return String representation of matches */
+  public String toString() {
```
dhruvi618, line 78: I'm not sure where we planning to use this (logging purposes or to send info to notifications part), but if we are planning to use this for notification, I feel like it might be helpful to use a comma delimiter for easy parsing on the receiver's end of the string? If this will only be used logging purposes then I think this is an intuitive way to format the string!

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,150 @@
...
+      participants.add(getParticipantFromEntity(entity));
+    }
+    return participants;
+  }
+
+  /** Remove Participant from datastore */
+  public void removeParticipant(String username) {
+    Key participantKey = KeyFactory.createKey(KIND_PARTICIPANT, username);
+    datastore.delete(participantKey);
+  }
```
dhruvi618, line 131: Since we are moving away from deleting participants and rather nulling out time/availability fields, will this method still need to be included?

backend/src/main/java/com/google/sps/data/Participant.java
kmh287: This sounds to me like a conflation of two data models that could be separated out in your datastore. One is a `User` model that stores any preferences the user opted to save and the other is a `Participant` model which represents a user actively looking for a match. Otherwise you risk Participant becoming like a Union class where it either represents someone looking for a match (fields not null) or just stored user information (nulled out match fields). WDYT?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -43,17 +37,6 @@
...
-  private static final String PROPERTY_USERNAME = "username";
-  private static final String PROPERTY_STARTTIMEAVAILABLE = "startTimeAvailable";
-  private static final String PROPERTY_ENDTIMEAVAILABLE = "endTimeAvailable";
-  private static final String PROPERTY_DURATION = "duration";
-  private static final String PROPERTY_TIMESTAMP = "timestamp";
-  private static final String PROPERTY_FIRSTPARTICIPANT = "firstParticipant";
-  private static final String PROPERTY_SECONDPARTICIPANT = "secondParticipant";
-
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
dhruvi618, line 41: I was wondering what your thoughts were on separating the functionality of the add-participant servlet? From my read of this method I see both it posting a participant to participant datastore and searching for an immediate match. I know we talked about this earlier when the datastores and matching algorithm were structured a little differently, but I was wondering whether or not there's still a benefit to trying to find an immediate match

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -40,41 +41,39 @@ public long addMatch(Match match) {
...
 
-  /** Return match based on match datastore key id */
-  @Nullable
-  public Match getMatchFromId(long matchId) {
-    Key matchKey = KeyFactory.createKey(KIND_MATCH, matchId);
-    try {
-      return getMatchFromEntity(datastore.get(matchKey));
-    } catch (EntityNotFoundException e) {
+  /** Return Match from entity, or null if entity is null */
+  private Match getMatchFromEntity(Entity entity) {
```
kmh287, line 49: tiny nit: I like having a method like this to convert from the datastore entity to the Plain Old Java Object (POJO) representation. Can it be static?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -57,40 +58,69 @@ public void addParticipant(Participant participant) {
...
   }
 
-  /** Update Participant with new matchId, null out availability to show currently matched */
-  public void updateNewMatch(String username, long matchId) {
+  /**
+   * Update Participant entity with username with new matchId
+   *
+   * @return true if valid username (entity found), false if not
+   */
+  public boolean updateMatchId(String username, long matchId) {
```
kmh287, line 77: Is the username this participant's username or the match's username?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -57,40 +58,69 @@ public void addParticipant(Participant participant) {
...
     Entity participantEntity = getEntityFromUsername(username);
+    if (participantEntity == null) {
+      return false;
```
kmh287, line 80: How does the application get into this state?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
kmh287, line 77: WDYT about removing this method and requiring callers to use `addParticipant` instead? The caller is then responsible for modifying the Participant model as needed and this code doesn't need to care about granular updates to specific properties.

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: I see. In our pod meeting today, we talked about this and decided to make this a new PR after this one.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: Removed this method.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: For logging purposes!

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: I think there definitely still is a benefit to trying to find an immediate match because it ensures that no two participants are ever compared multiple times.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: If `getParticipantFromUsername()` is called, since it returns `getParticipantFromEntity(getEntityFromUsername(username)`, and getEntityFromUsername(username) returns null if an entity is not found for that username.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: Participant's username

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: Done.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: Done.

backend/src/main/java/com/google/sps/data/Participant.java
kmh287: Would it make sense to author and submit the UserDatastore PR first?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -66,109 +49,96 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable);
-      participantEntity.setProperty(PROPERTY_DURATION, duration);
-      participantEntity.setProperty(PROPERTY_TIMESTAMP, timestamp);
-      datastore.put(participantEntity);
+      // Match not found, add participant to datastore
+      participantDatastore.addParticipant(newParticipant);
     }
 
     // Redirect back to the HTML page
     response.sendRedirect("/index.html");
```
dhruvi618, line 113: Since we are checking if there is an immediate match or not, I think we should also return the resulting match status to the frontend

jzwang8: When an immediate match is found, the participant doesn't have to be immediately notified since the javascript part will search for the match every 5 seconds anyways.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,43 +16,54 @@
...
 
   /**
-   * Find match of new participant or add to list of participants, return whether or not match was
-   * found right after being added
+   * @return Match of new participant with unmatched participants by comparing duration and
+   *     availibility, or null if no match yet
    */
-  public Match findMatch(List<Participant> participants, Participant newParticipant) {
+  @Nullable
+  public Match findMatch(Participant newParticipant) {
```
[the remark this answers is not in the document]
jzwang8: I moved add participant to datastore functionality to the AddParticipantServlet, but in this old commit, I actually added the participant to the datastore only after getting a list of unmatched participants from the datastore so that a participant would never be matched with themselves.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -37,18 +33,11 @@
...
-  // Datastore Key/Property constants
-  private static final String KEY_MATCH = "Match";
-  private static final String PROPERTY_FIRSTPARTICIPANT = "firstParticipant";
-  private static final String PROPERTY_SECONDPARTICIPANT = "secondParticipant";
-  private static final String PROPERTY_DURATION = "duration";
-  private static final String PROPERTY_TIMESTAMP = "timestamp";
-
   // JSON key constants
-  private static final String KEY_MATCHSTATUS = "matchStatus";
-  private static final String KEY_FIRSTPARTICIPANTUSERNAME = "firstParticipantUsername";
```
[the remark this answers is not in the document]
jzwang8: Yep, I actually just wanted to rename them since "key" was misleading since a Key is a datastore term.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -61,47 +50,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      int duration = (int) entity.getProperty(PROPERTY_DURATION);
-      long timestamp = (long) entity.getProperty(PROPERTY_TIMESTAMP);
-      Match match = new Match(id, firstParticipant, secondParticipant, duration, timestamp);
-      matches.add(match);
+    // Find participant's current match, if exists and not returned yet
+    Participant participant = participantDatastore.getParticipantFromUsername(username);
+    if (participant == null) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
+          "Participant with username " + username + "is invalid.");
```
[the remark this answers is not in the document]
jzwang8: Changed wording of the comment. This case really should never happen since all participants are added to the datastore in this PR. However, I'm changing this in another PR where I'll have Users vs Participants, in which case your suggestion would be more crucial.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Put participant in datastore. Overwrite participant entity if participant with same username
+   * already exists.
+   */
+  public void addParticipant(Participant participant) {
+    // Set properties of entity based on participant fields
+    Entity participantEntity = new Entity(KIND_PARTICIPANT, participant.getUsername());
+    participantEntity.setProperty(PROPERTY_USERNAME, participant.getUsername());
```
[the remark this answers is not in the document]
jzwang8: What do you mean? setProperty is an Entity method and only requires one line. Also, when you set an entity's property, it doesn't actually change the entity in the datastore. To change an entity that's in the datastore, you have to "put" the changed entity in to overwrite the entity.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /** Put Match in datastore and return the match key id */
+  public long addMatch(Match match) {
+    // Set properties of entity
+    Entity matchEntity = new Entity(KIND_MATCH);
+    matchEntity.setProperty(PROPERTY_FIRSTPARTICIPANTUSERNAME, match.getFirstParticipantUsername());
+    matchEntity.setProperty(
+        PROPERTY_SECONDPARTICIPANTUSERNAME, match.getSecondParticipantUsername());
+    matchEntity.setProperty(PROPERTY_DURATION, match.getDuration());
+    matchEntity.setProperty(PROPERTY_TIMESTAMP, match.getTimestamp());
```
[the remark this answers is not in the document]
jzwang8: Same response as your other comment! 😃

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: @TcgVanguardTroll Yep!

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: I feel like if we are searching for a match immediately then it might be worth sending back the resulting match immediately as well instead of waiting to return and traversing a list of all matches and then returning it

tcwanggoogle:
Have we rebase the changes on top of the master? I see many files that are already on master but still in this PR, resulting in 31 files. It'll be better if we can rebase the branch on top of master. (It looks like you pull in the commits from the master, and trying to merge those commits back to where the old master is. For example, IIUC, the current situation is something like:
```
F (jan-dev)
E
D (new_master)
C
B
A 
old_master
```
And your current PR has all A-F, which has a lot of files and changes that are already on remote master. But actually you want to rebase it correctly, so it should only show changes from E and F in this PR.

tcwanggoogle:
> Have we rebase the changes on top of the master? I see many files that are already on master but still in this PR, resulting in 31 files. It'll be better if we can rebase the branch on top of master. (It looks like you pull in the commits from the master, and trying to merge those commits back to where the old master is. For example, IIUC, the current situation is something like:
> 
> ```
> F (jan-dev)
> E
> D (new_master)
> C
> B
> A 
> old_master
> ```
> 
> And your current PR has all A-F, which has a lot of files and changes that are already on remote master. But actually you want to rebase it correctly, so it should only show changes from E and F in this PR.

Continuing on the example above, (and continue to guess the reason) if rebasing is not performed, but only merging is, it might result in something like
```
F (jan-dev)
D
C
E
B
A 
old_master
```
which essentially is the same thing, but just different ordering of commits, so the PR will have to show the diff is A-F than master, because it doesn't know A,B,C,E,D,F is the same as A,B,C,D,E,F.

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,48 +1,45 @@
...
 
-  /** Datastore ID */
-  private final long id;
   /** Google username (ldap) */
   private final String username;
-  /** Time user is starts being available */
-  private final ZonedDateTime startTimeAvailable;
+  /** Time user starts being available */
+  // TODO: make these fields final once Participant and User classes are created
+  private ZonedDateTime startTimeAvailable;
```
jzwang8, line 14: Change to final since Participant fields shouldn't change anymore

PR #25, opened 2020-07-10
bisection: Ignore directories that don't have any source code

tcwanggoogle:
Add a list of directories that the program needs to ignore, mainly
node_modules, as it might appear on user's local directory but
actually doesn't contain any source code

### DIFF
--- a/bisection/application.py
+++ b/bisection/application.py
@@ -24,6 +24,7 @@ COMMENT_STYLES = {
     '.xml': [('<!--', '-->')],
 }
 
+IGNORE_DIRECTORIES = ['node_modules', 'build']
 
 def count_comments_from_contents(contents, style):
     lines = 0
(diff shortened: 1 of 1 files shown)


bisection/application.py
```diff
@@ -74,6 +75,9 @@ def collect_programs(threshold_lines):
             os.path.join(directory, '../backend')
     ]:
         for root, _, files in os.walk(top):
+            # Ignore all ignored directories
+            if [x for x in IGNORE_DIRECTORIES if x in root]:
```
kmh287, line 79: Can you make IGNORE_DIRECTORIES a set and then check if root in IGNORE_DIRECTORIES?

tcwanggoogle: `root` can be something like "frontend/node_modules/xx", so I have to check if an entry in `IGNORE_DIRECTORIES` is a substring of `root` or not.

kmh287: ahh got it, thanks

PR #30, opened 2020-07-13
Add User and UserDatastore classes

jzwang8:
This PR introduces the User and UserDatastore classes. I'm planning on rebasing jan-dev on these changes so I can incorporate the split of the previous Participant class to two separate classes: User and Participant.

Files added:
- User.java: Added a User class that currently just has a User's username. In the future, this class will include preferences.

- UserDatastore.java: Added a UserDatastore class that stores Users to the datastore. In the future, this class will store each User's preferences and past matches.

### DIFF
--- a/backend/src/main/java/com/google/sps/data/User.java
+++ b/backend/src/main/java/com/google/sps/data/User.java
@@ -1,0 +1,28 @@
+package com.google.sps.data;
+
+
+/** A user with saved preferences. */
+public final class User {
+
+  /** Datastore ID */
+  private final long id;
+  /** Google username (ldap) */
+  private final String username;
+  /** Time of submitted form */
+  private final long timestamp;
+
+  /** Initialize constructor fields */
+  public Participant(long id, String username, long timestamp) {
+    this.id = id;
+    this.username = username;
+    this.timestamp = timestamp;
+  }
+
+  public long getId() {
+    return id;
+  }
+
+  public String getUsername() {
+    return username;
+  }
+}
--- a/backend/src/main/java/com/google/sps/data/User.java
+++ b/backend/src/main/java/com/google/sps/data/User.java
@@ -22,4 +22,8 @@ public final class User {
   public String getUsername() {
     return username;
   }
+
+  public String toString() {
+    return "username=" + username + "\n";
+  }
 }
--- a/backend/src/main/java/com/google/sps/data/User.java
+++ b/backend/src/main/java/com/google/sps/data/User.java
@@ -1,5 +1,7 @@
 package com.google.sps.data;
 
+import com.google.common.base.MoreObjects;
+
 /** A user with saved preferences. */
 public final class User {
 
@@ -24,6 +26,6 @@ public final class User {
   }
 
   public String toString() {
-    return "username=" + username + "\n";
+    return MoreObjects.toStringHelper(this).add("id", id).add("username", username).toString();
   }
 }
--- a/backend/src/main/java/com/google/sps/data/User.java
+++ b/backend/src/main/java/com/google/sps/data/User.java
@@ -1,6 +1,5 @@
 package com.google.sps.data;
 
-
 /** A user with saved preferences. */
 public final class User {
 
@@ -8,14 +7,12 @@ public final class User {
   private final long id;
   /** Google username (ldap) */
   private final String username;
-  /** Time of submitted form */
-  private final long timestamp;
 
   /** Initialize constructor fields */
-  public Participant(long id, String username, long timestamp) {
+  public User(long id, String username) {
+    // TODO(#32): add User interests
     this.id = id;
     this.username = username;
-    this.timestamp = timestamp;
   }
 
   public long getId() {
--- a/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
+++ b/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
@@ -46,12 +46,6 @@ public final class UserDatastore {
     }
   }
 
-  /** Return User from username, or null if user with username not in datastore */
-  @Nullable
-  public User getUserFromUsername(String username) {
-    return getUserFromEntity(getEntityFromUsername(username));
-  }
-
   /** Return user object from datastore user entity, or null if entity is null */
   @Nullable
   private static User getUserFromEntity(Entity entity) {
@@ -67,6 +61,11 @@ public final class UserDatastore {
     return new User(id, username);
   }
 
+  /** Return User from username, or null if user with username not in datastore */
+  public User getUserFromUsername(String username) {
+    return getUserFromEntity(getEntityFromUsername(username));
+  }
+
   /** Return String representation of users for logging purposes */
   public String toString() {
     StringBuilder sb = new StringBuilder();
@@ -74,7 +73,7 @@ public final class UserDatastore {
     PreparedQuery results = datastore.prepare(query);
     for (Entity entity : results.asIterable()) {
       User user = getUserFromEntity(entity);
-      sb.append("username=" + user.getUsername());
+      sb.append("username=" + user.getUsername() + "\n");
     }
     return sb.toString();
   }
--- a/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
+++ b/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
@@ -1,0 +1,48 @@
+package com.google.sps.datastore;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.Entity;
+import com.google.sps.data.User;
+
+/** Separates datastore method calls involving User type from caller */
+public final class UserDatastore {
+
+  // Datastore Key/Property constants
+  private static final String KIND_USER = "User";
+  private static final String PROPERTY_USERNAME = "username";
+  private static final String PROPERTY_TIMESTAMP = "timestamp";
+
+  /** Formatter for converting between String and ZonedDateTime */
+  private static final DateTimeFormatter formatter = DateTimeFormatter.ISO_ZONED_DATE_TIME;
+
+  /** Datastore */
+  private final DatastoreService datastore;
+
+  /** Constructor that takes in DatastoreService */
+  public UserDatastore(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  /** Put user in datastore. Overwrite user entity if user with same username already exists. */
+  public void addUser(User user) {
+    // Set properties of entity based on user fields
+    Entity userEntity = new Entity(KIND_USER, user.getUsername());
+    userEntity.setProperty(PROPERTY_USERNAME, user.getUsername());
+    userEntity.setProperty(PROPERTY_TIMESTAMP, user.getTimestamp());
+
+    // Insert entity into datastore
+    datastore.put(userEntity);
+  }
+
+  /** Return String representation of participants for logging purposes */
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    Query query = new Query(KIND_USER);
+    PreparedQuery results = datastore.prepare(query);
+    for (Entity entity : results.asIterable()) {
+      User user = getParticipantFromEntity(entity);
+      sb.append("username=" + user.getUsername());
+    }
+    return sb.toString();
+  }
+}
--- a/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
+++ b/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
@@ -60,6 +60,7 @@ public final class UserDatastore {
     }
 
     // Get entity properties
+    long id = (long) entity.getKey().getId();
     String username = (String) entity.getProperty(PROPERTY_USERNAME);
 
     // Create and return new User
--- a/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
+++ b/backend/src/main/java/com/google/sps/datastore/UserDatastore.java
@@ -10,10 +10,6 @@ public final class UserDatastore {
   // Datastore Key/Property constants
   private static final String KIND_USER = "User";
   private static final String PROPERTY_USERNAME = "username";
-  private static final String PROPERTY_TIMESTAMP = "timestamp";
-
-  /** Formatter for converting between String and ZonedDateTime */
-  private static final DateTimeFormatter formatter = DateTimeFormatter.ISO_ZONED_DATE_TIME;
 
   /** Datastore */
   private final DatastoreService datastore;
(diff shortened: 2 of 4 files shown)


backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,28 @@
+package com.google.sps.data;
+
+
```
kmh287, line 3: really tiny nit: remove extra newline here

backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,28 @@
...
+/** A user with saved preferences. */
+public final class User {
+
+  /** Datastore ID */
+  private final long id;
+  /** Google username (ldap) */
+  private final String username;
+  /** Time of submitted form */
+  private final long timestamp;
```
kmh287, line 12: A user who chooses to save preferences with ad-lib will have multiple form timestamps. Does this make sense on this model?

backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,28 @@
+package com.google.sps.data;
+
+
+/** A user with saved preferences. */
+public final class User {
+
+  /** Datastore ID */
```
kmh287, line 7: Add a TODO + issue to add user interests to this model as well?

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,48 @@
...
+/** Separates datastore method calls involving User type from caller */
+public final class UserDatastore {
+
+  // Datastore Key/Property constants
+  private static final String KIND_USER = "User";
+  private static final String PROPERTY_USERNAME = "username";
+  private static final String PROPERTY_TIMESTAMP = "timestamp";
+
+  /** Formatter for converting between String and ZonedDateTime */
+  private static final DateTimeFormatter formatter = DateTimeFormatter.ISO_ZONED_DATE_TIME;
```
kmh287, line 16: Delete unused?

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,48 @@
...
+  public void addUser(User user) {
+    // Set properties of entity based on user fields
+    Entity userEntity = new Entity(KIND_USER, user.getUsername());
+    userEntity.setProperty(PROPERTY_USERNAME, user.getUsername());
+    userEntity.setProperty(PROPERTY_TIMESTAMP, user.getTimestamp());
+
+    // Insert entity into datastore
+    datastore.put(userEntity);
+  }
+
```
kmh287, line 36: you'll also want a method to get users from the database, likely keyed by usernsme

backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,28 @@
...
+  private final long id;
+  /** Google username (ldap) */
+  private final String username;
+  /** Time of submitted form */
+  private final long timestamp;
+
+  /** Initialize constructor fields */
+  public Participant(long id, String username, long timestamp) {
```
kmh287, line 15: User

backend/src/main/java/com/google/sps/data/User.java
jzwang8: Added TODO and created an issue.

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
jzwang8: Got it, added. I wasn't sure if I should add them yet since I don't know how exactly I'll be using them.

jzwang8:
> Overall looks good. Be sure to add tests for the datastore class as well.

I've been adding datastore tests in FindMatchQueryTest. Do you think that's sufficient or should I add a DatastoreTest class, or a UserDatastoreTest class?

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -28,19 +24,49 @@ public void addUser(User user) {
...
+  /** Return User from username, or null if user with username not in datastore */
+  @Nullable
+  public User getUserFromUsername(String username) {
+    return getUserFromEntity(getEntityFromUsername(username));
+  }
+
+  /** Return user object from datastore user entity, or null if entity is null */
+  @Nullable
+  private static User getUserFromEntity(Entity entity) {
```
kmh287, line 57: I would make entity @NonNull here to avoid the null check and make the caller check.

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,80 @@
...
+    Entity userEntity = new Entity(KIND_USER, user.getUsername());
+    userEntity.setProperty(PROPERTY_USERNAME, user.getUsername());
+
+    // Insert entity into datastore
+    datastore.put(userEntity);
+  }
+
+  /** Return User Entity from username, or null if entity is not found */
+  @Nullable
+  private Entity getEntityFromUsername(String username) {
```
kmh287, line 40: tiny nit: 'from' suggests that the entity is extracted somehow from the String. I would recommend `getEntityForUsername` or even just `getEntity`

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Test
+  public void addOneUser() {
+    // Add one user to datastore
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    UserDatastore userDatastore = new UserDatastore(datastore);
+
+    User userA = new User(ID_DEFAULT, PERSON_A);
```
kmh287, line 61: tiny nit: this is all part of the `arrange` part of the test so they should be grouped together without newlines. Here and below

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Test
+  public void getNonexistentUser() {
+    // Try to get user from username that's not in datastore
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    UserDatastore userDatastore = new UserDatastore(datastore);
+
+    assertThat(userDatastore.getUserFromUsername(PERSON_A)).isEqualTo(null);
```
kmh287, line 110: isNull()

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -48,7 +49,7 @@ private Entity getEntityFromUsername(String username) {
 
   /** Return user object from datastore user entity, or null if entity is null */
   @Nullable
-  private static User getUserFromEntity(Entity entity) {
+  private static User getUserFromEntity(@Nonnull Entity entity) {
     if (entity == null) {
       return null;
     }
```
kmh287: this is now the caller's responsibility

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  public void addOneUser() {
+    // Add one user to datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    UserDatastore userDatastore = new UserDatastore(datastore);
+    User userA = new User(ID_DEFAULT, PERSON_A);
+    userDatastore.addUser(userA);
+
+    String expected = "username=" + PERSON_A + "\n";
+    assertThat(userDatastore.toString()).isEqualTo(expected);
```
kmh287, line 63: here and below: instead of checking userDataStore.toString, I would recommend accessing `datastore` (the DatasotreService) and manually checking that the expected inserts occurred

jzwang8: Got it, changed.

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -91,7 +104,7 @@ public void addGetTwoUsers() {
     assertThat(userDatastore.getUserFromUsername(PERSON_B).getUsername()).isEqualTo(PERSON_B);
   }
 
-  @Test
+  @Test(expected = NullPointerException.class)
```
kmh287, line 107: I would expect this test method not to throw. Where is the NPE coming from?

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -0,0 +1,115 @@
...
+    helper.tearDown();
+  }
+
+  @Test
+  public void addOneUser() throws EntityNotFoundException {
+    // Add one user to datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    UserDatastore userDatastore = new UserDatastore(datastore);
+    User userA = new User(ID_DEFAULT, PERSON_A);
+    userDatastore.addUser(userA);
```
kmh287, line 68: tiny nit: this is part of the `act` block of this test and should be separated from the code above and below by blank lines. Same comment on the test right below this one.

jzwang8: I'm not sure I understand what you mean. What's the code above line 65?

backend/src/test/java/com/google/sps/UserDatastoreTest.java
jzwang8: Since I got rid of the null check in `getUserFromEntity()` and made the parameter @Nonnull, since I'm deliberately making that argument null in getUserFromUsername(), a NullPointerException is thrown. How do I go about changing this?

kmh287: use a non-null username that won't be in the datastore

backend/src/test/java/com/google/sps/UserDatastoreTest.java
kmh287: Every unit test can be broken down into three distinct parts: arrange, act, and assert. Each of these blocks should be grouped together with no blank lines within them and with a blank line separating each of them. So two total blank lines per test. 

```
    UserDatastore userDatastore = new UserDatastore(datastore);
    User userA = new User(ID_DEFAULT, PERSON_A);
```

This is the arrange block since it sets up the DatastoreService, UserDatastore, and User objects for the test.

```
 userDatastore.addUser(userA);
```

Is the act block since this test is testing the behavior of addUser. 

```
    Key keyA = KeyFactory.createKey(KIND_USER, PERSON_A);
    Entity entityA = datastore.get(keyA);
    assertThat((String) entityA.getProperty(PROPERTY_USERNAME)).isEqualTo(PERSON_A);
```

This is the assert block since it's verifying that the call in act behaved as expected.

jzwang8: I see, thanks for the explanation. Done.

backend/src/test/java/com/google/sps/UserDatastoreTest.java
jzwang8: I did add a non-null String username (PERSON_A) that's not in the datastore. getEntity(PERSON_A) returns null, but getUserFromEntity() only accepts non-null entities.

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /** Return String representation of users for logging purposes */
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    Query query = new Query(KIND_USER);
+    PreparedQuery results = datastore.prepare(query);
+    for (Entity entity : results.asIterable()) {
+      User user = getUserFromEntity(entity);
+      sb.append("username=" + user.getUsername() + "\n");
```
tcwanggoogle, line 74: Make `"username=" + user.getUsername() + "\n"` the return value of a function of User and give it some generic name like user.toString(). So we can expand the function to log more fields in the future.

jzwang8: Got it, done!

backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,29 @@
...
+  public long getId() {
+    return id;
+  }
+
+  public String getUsername() {
+    return username;
+  }
+
+  public String toString() {
+    return "username=" + username + "\n";
```
kmh287, line 27: https://guava.dev/releases/20.0/api/docs/com/google/common/base/MoreObjects.ToStringHelper.html is a commonly used class for removing some of the extra boilerplate. If you use that then it shouldn't be too difficult to add the id as well.

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,81 @@
...
+    // Get entity properties
+    long id = (long) entity.getKey().getId();
+    String username = (String) entity.getProperty(PROPERTY_USERNAME);
+
+    // Create and return new User
+    return new User(id, username);
+  }
+
+  /** Return User from username, or null if user with username not in datastore */
+  public User getUserFromUsername(String username) {
```
kmh287, line 62: @Nullable

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -0,0 +1,81 @@
...
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    Query query = new Query(KIND_USER);
+    PreparedQuery results = datastore.prepare(query);
+    for (Entity entity : results.asIterable()) {
+      User user = getUserFromEntity(entity);
+      sb.append(user.toString());
+    }
+    return sb.toString();
+  }
```
kmh287, line 80: WDYT about reverting this before submitting this PR. The toString method isn't expected to run for a long time and a reasonable caller wouldn't expect toString to initiate a call to a database. Also reading from the database could become very very slow as the number of users grows.

backend/src/test/java/com/google/sps/UserDatastoreTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Before
+  public void setUp() {
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
```
kmh287, line 60: Out of scope for this PR but it would be nice to either find a corresponding JUnit Rule for this or just write our own. A JUnit Rule encapsulates some setUp and tearDown work so that it can be reused across tests without repeating the same code multiple times. 

Since you'll no doubt use JUnit in the future if you continue working with Java I'd recommend filing an issue and investigating.

PR #31, opened 2020-07-13
Connect frontend and backend via POST request containing form inputs

dhruvi618:
Files changed: 
- Form.js: Add Axios POST request with form input information
- PreferencesCheckbox.js: change type of defaultChecked value from string to boolean
- AddParticipantServlet.java: Parse HTTP request body using JSON objects, match/convert Javascript inputs to corresponding Java fields, and validate inputs.
- pom.xml: Add JSON dependency
- package.json: Add eslint and Axios dependency
- yarn.lock: Add axios dependencies and update outdated dependencies/plugins
- .gitignore: modify entry for `build/` to `frontend/build/`


backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
+    StringBuffer requestBuffer = new StringBuffer();
+    try {
+      BufferedReader reader = request.getReader();
+      String currentLine;
+      while ((currentLine = reader.readLine()) != null) {
+        requestBuffer.append(currentLine);
+      }
+    } catch (Exception e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not read request body");
     }
```
kmh287, line 72: Looks like a good candidate for extraction into a helper method

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
+
+    // Retrieve body of HTTP request
+    StringBuffer requestBuffer = new StringBuffer();
+    try {
+      BufferedReader reader = request.getReader();
+      String currentLine;
+      while ((currentLine = reader.readLine()) != null) {
+        requestBuffer.append(currentLine);
+      }
+    } catch (Exception e) {
```
kmh287, line 70: When catching exceptions you want to try to catch the most specific exception possible. Catching `Exception` is very broad and can lead to unexpected scenarios.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not read request body");
     }
-    String username = email.split("@")[0];
+    JSONObject obj = new JSONObject(requestBuffer.toString());
+
+    // Retrieve the timeAvailableUntil input and convert to a UTC ZonedDateTime
+    long timeAvailableUntil = obj.getJSONObject("formDetails").getLong("timeAvailableUntil");
```
kmh287, line 76: Please extract these Strings to constants like the ones at the top of this file. Doing so makes them easier to rename if the keys on the JSON need to be updated.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
-    // Request parameter values
-    // TODO: Check input values are filled (before allowing to submit)
-    UserService userService = UserServiceFactory.getUserService();
-    String email = userService.getCurrentUser().getEmail();
-    if (email == null) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
-      return;
+
+    // Retrieve body of HTTP request
+    StringBuffer requestBuffer = new StringBuffer();
```
kmh287, line 63: Prefer to use StringBuilder instead of StringBuiffer except when the object needs to be synchronized across threads.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
     if (duration <= 0) {
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid duration.");
       return;
     }
 
-    long timestamp = System.currentTimeMillis();
+    String role = obj.getJSONObject("formDetails").getString("role");
+    String productArea = obj.getJSONObject("formDetails").getString("productArea");
+    boolean savePreference = obj.getJSONObject("formDetails").getBoolean("savePreference");
+    String matchPreference = obj.getJSONObject("formDetails").getString("matchPreference");
```
kmh287, line 91: Please extract String constants

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
-
-    int duration = convertToPositiveInt(request.getParameter("duration"));
+    int duration = obj.getJSONObject("formDetails").getInt("duration");
     if (duration <= 0) {
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid duration.");
       return;
     }
 
-    long timestamp = System.currentTimeMillis();
+    String role = obj.getJSONObject("formDetails").getString("role");
```
kmh287, line 88: obj.getJSONObject("fromDetails") could be extracted to a local variable here since it's referenced on four lines consecutively

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -103,15 +120,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable);
+      participantEntity.setProperty(PROPERTY_STARTTIMEAVAILABLE, startTimeAvailable.toString());
+      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable.toString());
       participantEntity.setProperty(PROPERTY_DURATION, duration);
       participantEntity.setProperty(PROPERTY_TIMESTAMP, timestamp);
       datastore.put(participantEntity);
     }
 
-    // Redirect back to the HTML page
-    response.sendRedirect("/index.html");
```
kmh287, line 114: no more redirection?

frontend/yarn.lock
```diff
@@ -2018,7 +2018,7 @@ acorn@^6.0.1, acorn@^6.0.4, acorn@^6.2.1:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-6.4.1.tgz#531e58ba3f51b9dacb9a6646ca4debf5b14ca474"
   integrity sha512-ZVA9k326Nwrj3Cj9jlh3wGFutC2ZornPNARZwsNYqQYgN0EsV2d53w5RN/co65Ohn4sUAUtb1rSUAOD6XN9idA==
 
-acorn@^7.1.1:
+acorn@^7.1.1, acorn@^7.2.0:
```
kmh287, line 2021: Looks like a lot of dependencies were added to this file? Are these all transitive deps of axios?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
tcwanggoogle: +1

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,33 +58,48 @@
...
+      BufferedReader reader = request.getReader();
+      String currentLine;
+      while ((currentLine = reader.readLine()) != null) {
+        requestBuffer.append(currentLine);
+      }
+    } catch (Exception e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not read request body");
     }
-    String username = email.split("@")[0];
+    JSONObject obj = new JSONObject(requestBuffer.toString());
```
tcwanggoogle, line 73: Looks like extracting information from this JSONObject can be in its own function, so it's easier to extend the fields. This doPost can be just taking results of that function call to create a new Participant. IMO, this doPost() should be the "main" function of adding a new participant, so it's the best to move as much as code out of it and instead use this function to call other functions.

frontend/src/components/PreferencesCheckbox.js
```diff
@@ -23,7 +23,7 @@ export default function PreferencesCheckbox(props) {
           onChange={(event) => props.onChange(event.target.checked)}
           name="checkbox"
           color="primary"
-          defaultChecked="true"
+          defaultChecked={true}
```
tcwanggoogle, line 26: Just curious: Any particular reason to make this change? To fix a bug, or to format?

dhruvi618: This change was to avoid any type mismatch as the default value would not be the string value but the boolean true.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8, line 76: *endTimeAvailable. Also, how is this input a long if the user only input a time without a date?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -103,15 +120,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       // Match not found, insert participant entity into datastore
       Entity participantEntity = new Entity(KEY_PARTICIPANT);
       participantEntity.setProperty(PROPERTY_USERNAME, username);
-      participantEntity.setProperty(PROPERTY_STARTTIMEAVAILABLE, startTimeAvailable);
-      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable);
+      participantEntity.setProperty(PROPERTY_STARTTIMEAVAILABLE, startTimeAvailable.toString());
```
jzwang8, line 123: Not sure if changing it here it necessary since it's changed in my "Create Match and Participant datastores" PR, but I think to convert ZonedDateTimes to strings, you have to use the DateTimeFormatter.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -103,15 +120,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable.toString());
       participantEntity.setProperty(PROPERTY_DURATION, duration);
       participantEntity.setProperty(PROPERTY_TIMESTAMP, timestamp);
       datastore.put(participantEntity);
     }
 
-    // Redirect back to the HTML page
-    response.sendRedirect("/index.html");
+    response.setContentType("text/plain;charset=UTF-8");
+    response.getWriter().println("Recieved form input details!");
```
jzwang8, line 131: Receive*

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: For the prototype, we're redirecting back to the main page, but for the MVP, I think we're redirecting to the waiting page.

frontend/src/components/Form.js
```diff
@@ -56,6 +57,35 @@ export default function Form() {
...
+   */
+  function handleFormSubmission(event) {
+    // TODO(#34): Validate input values (before allowing to submit)
+
+    // Override browser's default behvaior to execute POST request
+    event.preventDefault();
+
+    // Gather all form inputs into one object
+    const formDetails = {
+      timeAvailableUntil: timeAvailableUntil.getTime(),
```
jzwang8, line 71: Is there any way to check that getTime() returns that time on the current day?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: The redirection will occur from the frontend instead

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Changed to catching and IOException

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Done!

frontend/yarn.lock
dhruvi618: They're not all transitive deps of axios, but running the `yarn` command updates outdated deps/plugins as well. I just updated the PR description to include these changes as well

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Thank you for the suggestion, I moved the JSON object extraction into its own function.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -56,39 +56,35 @@
   private static final String PROPERTY_FIRSTPARTICIPANT = "firstParticipant";
   private static final String PROPERTY_SECONDPARTICIPANT = "secondParticipant";
 
+  // HTTP Request JSON key constants
+  private static final String KEY_TIMEAVAILABLEUNTIL = "timeAvailableUntil";
+  private static final String KEY_DURATION = "duration";
+  private static final String KEY_ROLE = "role";
+  private static final String KEY_PRODUCTAREA = "productArea";
+  private static final String KEY_SAVEPREFERENCE = "savePreference";
+  private static final String KEY_MATCHPREFERENCE = "matchPreference";
```
jzwang8, line 65: I know Key is technically the correct term for JSON keys, but I called them JSON_DURATION to differentiate between Datastore Keys.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: When I send it values over from the frontend, I use `getTime()` which allows me to retrieve the long

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Since I am using the UTC system default zoneID for ZonedTimeDate, its not requiring me to specify a formatter. This works for when the property is inputted into the datastore and when it is retrieved as well.

frontend/src/components/Form.js
dhruvi618: Since users will be in different timezones and the times all get converted to UTC on the backend, I am not sure if converting the times to the current day would help us

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Thanks for the suggestion, done!

frontend/src/components/Form.js
jzwang8: So what does getTime() return? I assumed getTime() would return a time without specifying the date. However, if getTime() supposedly returns a long (because timeAvailableUntil is a long in AddParticipantServlet), that means that time must be on some date. What is that date? Is it the current date in their timezone? That's what we would want. Milliseconds since epoch isn't dependent on timezone. So my question is: is there a way to check the value that getTime() returns?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: See my other comment below about getTime() on the frontend

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
   private static final String PROPERTY_FIRSTPARTICIPANT = "firstParticipant";
   private static final String PROPERTY_SECONDPARTICIPANT = "secondParticipant";
 
+  // HTTP Request JSON key constants
+  private static final String JSON_TIMEAVAILABLEUNTIL = "timeAvailableUntil";
```
kmh287, line 60: I would prefix these with REQUEST_ and optionally remove the JSON part as well. The pertinent information is that these are keys into the request.

kmh287, line 60: TIME_AVAILABLE_UNTIL

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
...
+  private static final String JSON_DURATION = "duration";
+  private static final String JSON_ROLE = "role";
+  private static final String JSON_PRODUCTAREA = "productArea";
```
kmh287, line 63: PRODUCT_AREA

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
...
+  private static final String JSON_SAVEPREFERENCE = "savePreference";
```
kmh287, line 64: SAVE_PREFERENCE

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
...
+  private static final String JSON_MATCHPREFERENCE = "matchPreference";
```
kmh287, line 65: MATCH_PREFERENCE

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
...
+
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Request parameter values
-    // TODO: Check input values are filled (before allowing to submit)
+    JSONObject formDetails = retrieveRequestBody(request, response).getJSONObject("formDetails");
```
kmh287, line 69: "formDetails" can also be pulled out to a constant

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -54,10 +56,39 @@
...
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid duration.");
+      return;
+    }
+
+    String role = formDetails.getString(JSON_ROLE);
+    String productArea = formDetails.getString(JSON_PRODUCTAREA);
+    boolean savePreference = formDetails.getBoolean(JSON_SAVEPREFERENCE);
+    String matchPreference = formDetails.getString(JSON_MATCHPREFERENCE);
+
+    Long timestamp = System.currentTimeMillis();
```
kmh287, line 89: Can this still be a `long` instead of `Long`? The former is a primitive and the latter is a more expensive Object.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -103,15 +116,30 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
       datastore.put(participantEntity);
     }
 
-    // Redirect back to the HTML page
-    response.sendRedirect("/index.html");
+    response.setContentType("text/plain;charset=UTF-8");
+    response.getWriter().println("Received form input details!");
+  }
+
+  private JSONObject retrieveRequestBody(HttpServletRequest request, HttpServletResponse response)
```
kmh287, line 130: WDYT about having this method only take the request and then throwing the IOException to the caller to handle with the response?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -127,12 +155,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-          (ZonedDateTime) entity.getProperty(PROPERTY_STARTTIMEAVAILABLE);
-      ZonedDateTime endTimeAvailable =
-          (ZonedDateTime) entity.getProperty(PROPERTY_ENDTIMEAVAILABLE);
-      int duration = (int) entity.getProperty(PROPERTY_DURATION);
-      long timestamp = (long) entity.getProperty(PROPERTY_TIMESTAMP);
+          (ZonedDateTime) ZonedDateTime.parse(startTimeAvailableString);
+      String endTimeAvailableString = (String) entity.getProperty(PROPERTY_ENDTIMEAVAILABLE);
+      ZonedDateTime endTimeAvailable = (ZonedDateTime) ZonedDateTime.parse(endTimeAvailableString);
+      int duration = ((Long) entity.getProperty(PROPERTY_DURATION)).intValue();
+      Long timestamp = (Long) entity.getProperty(PROPERTY_TIMESTAMP);
```
kmh287, line 164: long?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -103,15 +122,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       // Match not found, insert participant entity into datastore
       Entity participantEntity = new Entity(KEY_PARTICIPANT);
       participantEntity.setProperty(PROPERTY_USERNAME, username);
-      participantEntity.setProperty(PROPERTY_STARTTIMEAVAILABLE, startTimeAvailable);
-      participantEntity.setProperty(PROPERTY_ENDTIMEAVAILABLE, endTimeAvailable);
+      participantEntity.setProperty(PROPERTY_START_TIME_AVAILABLE, startTimeAvailable.toString());
+      participantEntity.setProperty(PROPERTY_END_TIME_AVAILABLE, endTimeAvailable.toString());
       participantEntity.setProperty(PROPERTY_DURATION, duration);
       participantEntity.setProperty(PROPERTY_TIMESTAMP, timestamp);
       datastore.put(participantEntity);
```
tcwanggoogle, line 129: Optional (and maybe beyond the scope of this PR): We can match the if part by creating a function `addParticipantToDatastore`, where we create an entity and add it to the datastore.

frontend/src/components/PreferencesCheckbox.js
tcwanggoogle: thanks for the explanation

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
tcwanggoogle: And we can pass the newParticipant to the function, to use the fields.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -47,51 +49,68 @@
...
     long timestamp = System.currentTimeMillis();
 
+    // Retrieve user email address via Users API and parse for ldap
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    if (email == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
+      return;
+    }
+    String username = email.split("@")[0];
```
tcwanggoogle, line 104: I think there's still an opportunity to create a function for this part, as something like `getInfoFromJSONObject`, so we can further reduce the body of `doPost`. I am OK with refactoring in another PR, but I'm worried this function getting much larger (and harder to understand) when we add more and more functionalities.

dhruvi618: Modularize this functionality into its own method, thanks!

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Refactored this code to be part of its own method `addParticipanyEntityToDatastore`

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -47,51 +49,64 @@
...
+  private static final String REQUEST_FORM_DETAILS = "formDetails";
+  private static final String REQUEST_TIME_AVAILABLE_UNTIL = "timeAvailableUntil";
+  private static final String REQUEST_DURATION = "duration";
+  private static final String REQUEST_ROLE = "role";
+  private static final String REQUEST_PRODUCT_AREA = "productArea";
+  private static final String REQUEST_SAVE_PREFERENCE = "savePreference";
+  private static final String REQUEST_MATCH_PREFERENCE = "matchPreference";
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
kmh287, line 69: Doesn't need to be in this PR but please file an issue to unit test this method.

frontend/src/components/Form.js
dhruvi618: getTime() returns the number representing the milliseconds elapsed between 1 January 1970 00:00:00 UTC and the given date. What check do you want to do on that value?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Thanks for the suggestion, done!

PR #46, opened 2020-07-15
NEW: Create MatchDatastore and ParticipantDatastore + unit tests

jzwang8:
Files changed:
- pom.xml: Add datastore unit testing dependencies
- Match.java: Change Participant fields to String usernames
- Participant.java: Add currentMatchID to keep track of match
- MatchDatastore.java: Datastore for Match kinds to add and get
- ParticipantDatastore.java: Datastore for Participant kinds to add and get
- MatchDatastoreTest.java: Unit tests for MatchDatastore
- ParticipantDatastoreTest.java: Unit tests for Participant Datastore
- FindMatchQuery.java: Pass Datastore into constructor instead of in method, filter datastore by duration instead of searching all participants and comparing duration
- FindMatchQueryTest.java: Fix syntax to correspond with changes in other files
- UserDatastoreTest.java: Change one of the tests
- AddParticipantServlet.java: Front end/back end input communication
- SearchMatchServlet.java: Search for matches based on saved matchId in a Participant

### DIFF
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -24,9 +24,7 @@ import java.util.List;
 /** Class used to find a match in a list of Participants with the most recently added Participant */
 public final class FindMatchQuery {
 
-  /** Maximum difference in duration to be compatible */
-  private static int MAX_DURATION_DIFF = 15;
-  /** extra padding time to ensure large enough meeting time block */
+  /** Extra padding time to ensure large enough meeting time block */
   private static int PADDING_TIME = 15;
   /** Reference clock */
   private Clock clock;
@@ -37,28 +35,24 @@ public final class FindMatchQuery {
   }
 
   /**
-   * Find match of new participant or add to list of participants, return whether or not match was
-   * found right after being added
+   * Find and return match of new participant with unmatched participants by comparing duration and
+   * availibility or return null if no match yet
    */
   public Match findMatch(ParticipantDatastore participantDatastore, Participant newParticipant) {
-    // TODO: filter participants here
-    List<Participant> participants = participantDatastore.getAllParticipants();
+    int newDuration = newParticipant.getDuration();
+    List<Participant> unmatchedParticipants =
+        participantDatastore.getUnmatchedParticipants(newDuration);
+
+    // Add newParticipant to datastore
+    participantDatastore.addParticipant(newParticipant);
 
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
-    System.out.println(dateTime);
 
-    // Compare new participant preferences with others in list to find match
-    for (Participant currParticipant : participants) {
-
-      // Check if participants are looking for similar meeting duration
-      int newDuration = newParticipant.getDuration();
+    // Compare new participant preferences with other participants with similar duration to find
+    // match
+    for (Participant currParticipant : unmatchedParticipants) {
       int currDuration = currParticipant.getDuration();
-      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
-
-      if (!compatibleDuration) {
-        continue;
-      }
       int duration = Math.min(newDuration, currDuration);
 
       // Check if participants are both free for that duration + extra
@@ -70,11 +64,10 @@ public final class FindMatchQuery {
           dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
 
       if (compatibleTime) {
-        // TODO: change match ID (currently -1 for easy error checking)
         return new Match(
             /* id= */ -1L,
-            newParticipant,
-            currParticipant,
+            participantDatastore.getParticipantFromUsername(newParticipant.getUsername()).getId(),
+            currParticipant.getId(),
             duration,
             dateTime.toInstant().toEpochMilli());
       }
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -47,7 +47,7 @@ public final class FindMatchQuery {
   @Nullable
   public Match findMatch(Participant newParticipant) {
     int duration = newParticipant.getDuration();
-
+    
     // Get list of unmatched participants with same duration as newParticipant
     List<Participant> sameDurationParticipants =
         participantDatastore.getSameDurationParticipants(duration);
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -28,7 +28,7 @@ public final class FindMatchQuery {
   /** Maximum difference in duration to be compatible */
   private static final int MAX_DURATION_DIFF = 15;
   /** Extra padding time to ensure large enough meeting time block */
-  private static final int PADDING_TIME = 15;
+  private static final int PADDING_TIME = 10;
   /** Reference clock */
   private final Clock clock;
   /** Datastore of Participants */
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -19,6 +19,7 @@ import com.google.sps.data.Participant;
 import com.google.sps.datastore.ParticipantDatastore;
 import java.time.Clock;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /** Class used to find a match for new Participant with unmatched Participants in datastore */
@@ -26,8 +27,6 @@ public final class FindMatchQuery {
 
   /** Extra padding time in minutes to ensure large enough meeting time block */
   private static final int PADDING_MINUTES = 10;
-  /** Conversion from minutes to milliseconds */
-  private static final int MINUTES_TO_MILLIS = 60000;
   /** Reference clock */
   private final Clock clock;
   /** Datastore of Participants */
@@ -62,7 +61,7 @@ public final class FindMatchQuery {
       long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
       long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
       boolean compatibleTime =
-          (currentTimeMillis + (duration + PADDING_MINUTES) * MINUTES_TO_MILLIS)
+          (currentTimeMillis + TimeUnit.MINUTES.toMillis(duration + PADDING_MINUTES))
               < earliestEndTimeAvailable;
 
       if (compatibleTime) {
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -21,7 +21,7 @@ import java.time.Clock;
 import java.time.ZonedDateTime;
 import java.util.List;
 
-/** Class used to find a match in a list of Participants with the most recently added Participant */
+/** Class used to find a match for new Participant with unmatched Participants in datastore */
 public final class FindMatchQuery {
 
   /** Maximum difference in duration to be compatible */
@@ -41,6 +41,7 @@ public final class FindMatchQuery {
    * availibility or return null if no match yet
    */
   public Match findMatch(ParticipantDatastore participantDatastore, Participant newParticipant) {
+    // Get list of unmatched participants in datastore
     List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
 
     // Add newParticipant to datastore
@@ -52,13 +53,13 @@ public final class FindMatchQuery {
     // Compare new participant preferences with other participants with similar duration to find
     // match
     for (Participant currParticipant : unmatchedParticipants) {
-      System.out.println(currParticipant.getUsername());
       // Check if participants are looking for similar meeting duration
       int newDuration = newParticipant.getDuration();
       int currDuration = currParticipant.getDuration();
       boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
 
       if (!compatibleDuration) {
+        // Not a match
         continue;
       }
       int duration = Math.min(newDuration, currDuration);
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -24,6 +24,8 @@ import java.util.List;
 /** Class used to find a match in a list of Participants with the most recently added Participant */
 public final class FindMatchQuery {
 
+  /** Maximum difference in duration to be compatible */
+  private static int MAX_DURATION_DIFF = 15;
   /** Extra padding time to ensure large enough meeting time block */
   private static int PADDING_TIME = 15;
   /** Reference clock */
@@ -39,9 +41,7 @@ public final class FindMatchQuery {
    * availibility or return null if no match yet
    */
   public Match findMatch(ParticipantDatastore participantDatastore, Participant newParticipant) {
-    int newDuration = newParticipant.getDuration();
-    List<Participant> unmatchedParticipants =
-        participantDatastore.getUnmatchedParticipants(newDuration);
+    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
 
     // Add newParticipant to datastore
     participantDatastore.addParticipant(newParticipant);
@@ -52,7 +52,15 @@ public final class FindMatchQuery {
     // Compare new participant preferences with other participants with similar duration to find
     // match
     for (Participant currParticipant : unmatchedParticipants) {
+      System.out.println(currParticipant.getUsername());
+      // Check if participants are looking for similar meeting duration
+      int newDuration = newParticipant.getDuration();
       int currDuration = currParticipant.getDuration();
+      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
+
+      if (!compatibleDuration) {
+        continue;
+      }
       int duration = Math.min(newDuration, currDuration);
 
       // Check if participants are both free for that duration + extra
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -46,28 +46,17 @@ public final class FindMatchQuery {
    */
   @Nullable
   public Match findMatch(Participant newParticipant) {
-    // Get list of unmatched participants in datastore
-    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
+    int duration = newParticipant.getDuration();
 
-    // Add newParticipant to datastore
-    participantDatastore.addParticipant(newParticipant);
+    // Get list of unmatched participants with same duration as newParticipant
+    List<Participant> sameDurationParticipants =
+        participantDatastore.getSameDurationParticipants(duration);
 
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
 
     // Compare new participant preferences with other participants to find match
-    for (Participant currParticipant : unmatchedParticipants) {
-      // Check if participants are looking for similar meeting duration
-      int newDuration = newParticipant.getDuration();
-      int currDuration = currParticipant.getDuration();
-      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
-
-      if (!compatibleDuration) {
-        // Not a match
-        continue;
-      }
-      int duration = Math.min(newDuration, currDuration);
-
+    for (Participant currParticipant : sameDurationParticipants) {
       // Check if participants are both free for that duration + extra
       ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
       ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -25,8 +25,6 @@ import javax.annotation.Nullable;
 /** Class used to find a match for new Participant with unmatched Participants in datastore */
 public final class FindMatchQuery {
 
-  /** Maximum difference in duration to be compatible */
-  private static final int MAX_DURATION_DIFF = 15;
   /** Extra padding time to ensure large enough meeting time block */
   private static final int PADDING_TIME = 10;
   /** Reference clock */
@@ -55,10 +53,11 @@ public final class FindMatchQuery {
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
 
+    ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
+
     // Compare new participant preferences with other participants to find match
     for (Participant currParticipant : sameDurationParticipants) {
       // Check if participants are both free for that duration + extra
-      ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
       ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
       ZonedDateTime earliestEndTimeAvailable =
           getEarlier(newEndTimeAvailable, currEndTimeAvailable);
@@ -66,6 +65,7 @@ public final class FindMatchQuery {
           dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
 
       if (compatibleTime) {
+        // Create and return match (id is irrelevant because it's assigned when put into datastore
         return new Match(
             /* id= */ -1L,
             newParticipant.getUsername(),
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -55,8 +55,7 @@ public final class FindMatchQuery {
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
 
-    // Compare new participant preferences with other participants with similar duration to find
-    // match
+    // Compare new participant preferences with other participants to find match
     for (Participant currParticipant : unmatchedParticipants) {
       // Check if participants are both free for that duration + extra
       ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -47,7 +47,11 @@ public final class FindMatchQuery {
   @Nullable
   public Match findMatch(Participant newParticipant) {
     int duration = newParticipant.getDuration();
+<<<<<<< HEAD
     
+=======
+
+>>>>>>> Change so that only participants with same duration can be matched
     // Get list of unmatched participants with same duration as newParticipant
     List<Participant> sameDurationParticipants =
         participantDatastore.getSameDurationParticipants(duration);
@@ -56,7 +60,11 @@ public final class FindMatchQuery {
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
 
     // Compare new participant preferences with other participants to find match
+<<<<<<< HEAD
     for (Participant currParticipant : unmatchedParticipants) {
+=======
+    for (Participant currParticipant : sameDurationParticipants) {
+>>>>>>> Change so that only participants with same duration can be matched
       // Check if participants are both free for that duration + extra
       ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
       ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -65,9 +65,8 @@ public final class FindMatchQuery {
           dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
 
       if (compatibleTime) {
-        // Create and return match (id is irrelevant because it's assigned when put into datastore
+        // Create and return match
         return new Match(
-            /* id= */ -1L,
             newParticipant.getUsername(),
             currParticipant.getUsername(),
             duration,
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -24,8 +24,8 @@ import javax.annotation.Nullable;
 /** Class used to find a match for new Participant with unmatched Participants in datastore */
 public final class FindMatchQuery {
 
-  /** Extra padding time to ensure large enough meeting time block */
-  private static final int PADDING_TIME = 10;
+  /** Extra padding time in minutes to ensure large enough meeting time block */
+  private static final int PADDING_MINUTES = 10;
   /** Conversion from minutes to milliseconds */
   private static final int MINUTES_TO_MILLIS = 60000;
   /** Reference clock */
@@ -62,7 +62,7 @@ public final class FindMatchQuery {
       long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
       long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
       boolean compatibleTime =
-          (currentTimeMillis + (duration + PADDING_TIME) * MINUTES_TO_MILLIS)
+          (currentTimeMillis + (duration + PADDING_MINUTES) * MINUTES_TO_MILLIS)
               < earliestEndTimeAvailable;
 
       if (compatibleTime) {
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -47,11 +47,7 @@ public final class FindMatchQuery {
   @Nullable
   public Match findMatch(Participant newParticipant) {
     int duration = newParticipant.getDuration();
-<<<<<<< HEAD
-    
-=======
 
->>>>>>> Change so that only participants with same duration can be matched
     // Get list of unmatched participants with same duration as newParticipant
     List<Participant> sameDurationParticipants =
         participantDatastore.getSameDurationParticipants(duration);
@@ -60,11 +56,7 @@ public final class FindMatchQuery {
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
 
     // Compare new participant preferences with other participants to find match
-<<<<<<< HEAD
-    for (Participant currParticipant : unmatchedParticipants) {
-=======
     for (Participant currParticipant : sameDurationParticipants) {
->>>>>>> Change so that only participants with same duration can be matched
       // Check if participants are both free for that duration + extra
       ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
       ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -54,8 +54,10 @@ public final class FindMatchQuery {
 
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
-    // Compare new participant preferences with other participants to find match
-    for (Participant currParticipant : sameDurationParticipants) {
+
+    // Compare new participant preferences with other participants with similar duration to find
+    // match
+    for (Participant currParticipant : unmatchedParticipants) {
       // Check if participants are both free for that duration + extra
       ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
       ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -54,7 +54,6 @@ public final class FindMatchQuery {
 
     // Set reference date time using clock
     ZonedDateTime dateTime = ZonedDateTime.now(clock);
-
     // Compare new participant preferences with other participants to find match
     for (Participant currParticipant : sameDurationParticipants) {
       // Check if participants are both free for that duration + extra
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -74,8 +74,11 @@ public final class FindMatchQuery {
       if (compatibleTime) {
         return new Match(
             /* id= */ -1L,
-            participantDatastore.getParticipantFromUsername(newParticipant.getUsername()).getId(),
-            currParticipant.getId(),
+            participantDatastore.getKeyFromId(
+                participantDatastore
+                    .getParticipantFromUsername(newParticipant.getUsername())
+                    .getId()),
+            participantDatastore.getKeyFromId(currParticipant.getId()),
             duration,
             dateTime.toInstant().toEpochMilli());
       }
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -18,7 +18,6 @@ import com.google.sps.data.Match;
 import com.google.sps.data.Participant;
 import com.google.sps.datastore.ParticipantDatastore;
 import java.time.Clock;
-import java.time.ZonedDateTime;
 import java.util.List;
 import javax.annotation.Nullable;
 
@@ -27,6 +26,8 @@ public final class FindMatchQuery {
 
   /** Extra padding time to ensure large enough meeting time block */
   private static final int PADDING_TIME = 10;
+  /** Conversion from minutes to milliseconds */
+  private static final int MINUTES_TO_MILLIS = 60000;
   /** Reference clock */
   private final Clock clock;
   /** Datastore of Participants */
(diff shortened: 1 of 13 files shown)


backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,48 +1,45 @@
...
+  /** Time user starts being available */
+  // TODO: make these fields final once Participant and User classes are created
+  private ZonedDateTime startTimeAvailable;
   /** Time user is available until */
-  private final ZonedDateTime endTimeAvailable;
+  private ZonedDateTime endTimeAvailable;
   /** How long user wants to chat */
-  private final int duration;
+  private int duration;
+  /** Current match in datastore, null if never found a match or already been returned */
```
tcwanggoogle, line 19: "null" -> "0"

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,48 +1,45 @@
...
   private final long timestamp;
 
   /** Initialize constructor fields */
   public Participant(
-      long id,
       String username,
       ZonedDateTime startTimeAvailable,
       ZonedDateTime endTimeAvailable,
       int duration,
+      long currentMatchId,
```
tcwanggoogle, line 30: Is it possible that when we create a Participant, we already know the match id? If not, should this field be always initialized to 0?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -82,11 +76,6 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         ZonedDateTime.ofInstant(Instant.ofEpochMilli(timeAvailableUntil), zoneId);
 
     int duration = formDetails.getInt(REQUEST_DURATION);
-    if (duration <= 0) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid duration.");
-      return;
-    }
-
```
tcwanggoogle, line 89: Removed by accident? I think we still need this validation?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -61,47 +50,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    // Check if match exists and not returned yet
+    if (currentMatchId == 0) {
+      // No match yet
+      JSONObject matchDoesNotExist = new JSONObject();
+      matchDoesNotExist.put(JSON_MATCHSTATUS, "false");
 
-        break;
+      // Send the JSON back as the response
+      response.setContentType("application/json");
+      response.getWriter().println(matchDoesNotExist.toString());
```
tcwanggoogle, line 76: optional: You can early return here, so we don't need `else` part.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,27 +88,49 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      currParticipant.setEndTimeAvailable(null);
+      currParticipant.setDuration(0);
+      participantDatastore.addParticipant(currParticipant);
+
+      // Update newParticipant entity with new currentMatchId and null availability
+      newParticipant.setCurrentMatchId(matchId);
+      newParticipant.setStartTimeAvailable(null);
+      newParticipant.setEndTimeAvailable(null);
+      newParticipant.setDuration(0);
+      participantDatastore.addParticipant(newParticipant);
```
tcwanggoogle, line 129: Can we encapsulate these into function calls?

backend/src/main/java/com/google/sps/data/Match.java
```diff
@@ -1,19 +1,16 @@
...
-  /** Datastore entity ID */
+  /** Match datastore key ID */
   private final long id;
-  /** First participant */
-  private final Participant firstParticipant;
-  /** Second participant */
-  private final Participant secondParticipant;
+  /** First participant username (datastore key name) */
+  private final String firstParticipantUsername;
+  /** Second participant username (datastore key name) */
```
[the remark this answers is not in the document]
kmh287: I recommended against storing the full Participant so that we could keep the data models fully separate. Otherwise the code that inserts Match objects into the datastore would have to know how to store and retrieve Participants as well.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,47 +16,47 @@
...
+/** Class used to find a match for new Participant with unmatched Participants in datastore */
 public final class FindMatchQuery {
 
   /** Maximum difference in duration to be compatible */
-  private static int MAX_DURATION_DIFF = 15;
-  /** extra padding time to ensure large enough meeting time block */
-  private static int PADDING_TIME = 15;
+  private static final int MAX_DURATION_DIFF = 15;
+  /** Extra padding time to ensure large enough meeting time block */
+  private static final int PADDING_TIME = 10;
```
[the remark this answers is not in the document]
jzwang8: Padding time is used to make sure there's enough time for participants to get ready and join the meet call for the duration before at least one of them is unavailable. I also had to update the tests because I had 2 tests testing edge cases.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,47 +16,47 @@
...
+  public Match findMatch(Participant newParticipant) {
+    int duration = newParticipant.getDuration();
 
-      // Check if participants are looking for similar meeting duration
-      int newDuration = newParticipant.getDuration();
-      int currDuration = currParticipant.getDuration();
-      boolean compatibleDuration = Math.abs(newDuration - currDuration) <= MAX_DURATION_DIFF;
+    // Get list of unmatched participants with same duration as newParticipant
+    List<Participant> sameDurationParticipants =
+        participantDatastore.getSameDurationParticipants(duration);
```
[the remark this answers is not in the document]
jzwang8: Yep, only matching with same duration. Deleted MAX_DURATION_DIFF, thanks!

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -66,11 +66,10 @@ public Match findMatch(List<Participant> participants, Participant newParticipan
           dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
 
       if (compatibleTime) {
-        // TODO: change match ID (currently -1 for easy error checking)
```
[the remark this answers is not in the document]
jzwang8: Match IDs can't be assigned, they correspond to the key of the item in the datastore.

backend/src/main/java/com/google/sps/data/Match.java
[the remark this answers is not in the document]
jzwang8: Less overhead, and you only need the username=key to find the whole participant in the datastore.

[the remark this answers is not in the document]
tcwanggoogle: Ok. If we only ever only need the username part, I have no problems. I was just raising the question because it seems the same amount of storage in terms of storing a pointer to an object (I hope this is the implementation in Java) vs a string. But you got valid points here.

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: When we create a participant just to pass to the datastore to add a participant, yes, matchId would be 0. However, when we return a Participant (from a username key that leads you to an entity whose properties are the Participant's fields), we definitely want to return the matchId in the datastore, so no need to always initialize to 0 if you'd also have to setMatchId() separately too.

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
tcwanggoogle: Does it mean we can't assign a match ID when creating a Match? If so, we should remove this field from the constructor and set the field to a default value. (Otherwise people who's using this constructor may think they have to assign a value for it)

backend/src/main/java/com/google/sps/data/Participant.java
tcwanggoogle: Got it. I don't know if Java supports default value for a field in a constructor. Or we should create a new constructor, or just update javadoc to let the user know that it's OK to assign 0 to this field if matchId is unknown during construction.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,27 +92,34 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      nullAvailabilityAdd(participantDatastore, currParticipant, matchId);
+      nullAvailabilityAdd(participantDatastore, newParticipant, matchId);
     } else {
-      datastore.put(addParticipantEntityToDatastore(newParticipant));
+      // Match not found, add participant to datastore
+      participantDatastore.addParticipant(newParticipant);
     }
-
-    response.setContentType("text/plain;charset=UTF-8");
-    response.getWriter().println("Received form input details!");
```
dhruvi618, line 122: WDYT about leaving these lines in the servlet for testing purposes just as a confirmation to the frontend that the form details were recieved?

jzwang8: Oops, didn't mean to delete that! Thanks for catching that!

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
jzwang8: Yes, we can't assign a match ID when creating a match. But by saving the match id, given a match, we can find the corresponding match in the datastore. Looking at the rest of the code, it doesn't look like I ever really need to do that, so I can take the matchId out of a Match object.

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: I'll add a javadoc!

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,48 +1,45 @@
...
       ZonedDateTime endTimeAvailable,
       int duration,
+      long matchId,
       long timestamp) {
-    this.id = id;
     this.username = username;
-
     Preconditions.checkArgument(
         startTimeAvailable.isBefore(endTimeAvailable),
         "Start available time must be before end available time.");
```
dhruvi618, line 35: Just wondering, what happens if this precondition fails?

jzwang8: Good question, I just checked and it throws an IllegalArgumentException. https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

jzwang8: IllegalArgumentException/mvn test error

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,65 +1,91 @@
...
+  }
+
+  public long getEndTimeAvailable() {
     return endTimeAvailable;
   }
 
+  public void setEndTimeAvailable(long endTimeAvailable) {
+    this.endTimeAvailable = endTimeAvailable;
+  }
+
```
kmh287, line 60: The introduction of setters to this class takes away the benefit of deep immutability. It's always a good idea to keep your model classes like this deeply immutable because deeply immutable models are inherently thread-safe. 

If the prod code doesn't need to change these values we should take out the setters. If we do have cases that need to mutate these values I'd recommend a slightly different approach:

```
public Participant withMatchId(long newMatchId) {
  return new Participant(
    userName,
    startTimeAvailable,
    endTimeAvailable,
    duration,
    newMatchId,
    timestamp);
}
```

Which allows you to maintain deep immutability while also making it easy to create new models with the desired values.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,72 @@
...
+
+    // Return match key id
+    return matchEntity.getKey().getId();
+  }
+
+  /** Return Match from entity, or null if entity is null */
+  private static Match getMatchFromEntity(Entity entity) {
+    if (entity == null) {
+      return null;
+    }
```
kmh287, line 50: Let's make entity @NonNull and require callers to null check. The benefit of that is that we can handle nulls where they're generated instead of defending against possible nulls downstream.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /** Put Match in datastore and return the match key id */
+  public long addMatch(Match match) {
+    // Set properties of entity
+    Entity matchEntity = new Entity(KIND_MATCH);
+    matchEntity.setProperty(PROPERTY_FIRSTPARTICIPANTUSERNAME, match.getFirstParticipantUsername());
+    matchEntity.setProperty(
+        PROPERTY_SECONDPARTICIPANTUSERNAME, match.getSecondParticipantUsername());
+    matchEntity.setProperty(PROPERTY_DURATION, match.getDuration());
+    matchEntity.setProperty(PROPERTY_TIMESTAMP, match.getTimestamp());
```
kmh287, line 37: optional: It might be nice to have a parallel method to getMatchFromEntity called getEntityFromMatch that encapsulates this logic

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,118 @@
...
+    long matchId = (long) entity.getProperty(PROPERTY_MATCHID);
+    long timestamp = (long) entity.getProperty(PROPERTY_TIMESTAMP);
+
+    // Create and return new Participant
+    return new Participant(
+        username, startTimeAvailable, endTimeAvailable, duration, matchId, timestamp);
+  }
+
+  /** Return list of all unmatched participants with same duration */
+  public List<Participant> getSameDurationParticipants(int duration) {
```
kmh287, line 96: By this part of the code we've lost the concept of "same-ness" i.e. this code doesn't know about the participant we're comparing to. How about getParticipantsWithDuration (and updated javadocs)?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,118 @@
...
+    for (Entity entity : results.asIterable()) {
+      participants.add(getParticipantFromEntity(entity));
+    }
+    return participants;
+  }
+
+  /** Remove Participant from datastore */
+  public void removeParticipant(String username) {
+    Key participantKey = KeyFactory.createKey(KIND_PARTICIPANT, username);
+    datastore.delete(participantKey);
```
kmh287, line 116: any exceptions we need to worry about here if the key isn't in the datastore when calling delete?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,23 +87,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      long matchId = matchDatastore.addMatch(match);
+
+      System.out.println(matchDatastore.getMatchFromId(matchId).toString());
+
+      // Update participant entities with new matchId and null availability
+      Participant currParticipant =
+          participantDatastore.getParticipantFromUsername(match.getSecondParticipantUsername());
+      nullAvailabilityAdd(participantDatastore, currParticipant, matchId);
+      nullAvailabilityAdd(participantDatastore, newParticipant, matchId);
     } else {
```
kmh287, line 116: Now that user data is kept separate is this still needed?

backend/src/test/java/com/google/sps/MatchDatastoreTest.java
```diff
@@ -0,0 +1,142 @@
...
+        .isEqualTo(PERSON_B);
+    assertThat(((Long) entity1.getProperty(PROPERTY_DURATION)).intValue())
+        .isEqualTo(DURATION_DEFAULT);
+    assertThat((long) entity1.getProperty(PROPERTY_TIMESTAMP)).isEqualTo(TIMESTAMP_DEFAULT);
+    assertThat((String) entity2.getProperty(PROPERTY_FIRSTPARTICIPANTUSERNAME)).isEqualTo(PERSON_C);
+    assertThat((String) entity2.getProperty(PROPERTY_SECONDPARTICIPANTUSERNAME))
+        .isEqualTo(PERSON_D);
+    assertThat(((Long) entity2.getProperty(PROPERTY_DURATION)).intValue())
+        .isEqualTo(DURATION_DEFAULT);
+    assertThat((long) entity2.getProperty(PROPERTY_TIMESTAMP)).isEqualTo(TIMESTAMP_DEFAULT);
```
kmh287, line 114: Can you check if the casts are necessary? I suspect they can be removed

backend/src/test/java/com/google/sps/ParticipantDatastoreTest.java
```diff
@@ -0,0 +1,219 @@
...
+    Entity entity = datastore.get(key);
+    assertThat((String) entity.getProperty(PROPERTY_USERNAME)).isEqualTo(PERSON_A);
+    assertThat((long) entity.getProperty(PROPERTY_STARTTIMEAVAILABLE))
+        .isEqualTo(STARTTIMEAVAILABLE_DEFAULT);
+    assertThat((long) entity.getProperty(PROPERTY_ENDTIMEAVAILABLE))
+        .isEqualTo(ENDTIMEAVAILABLE_DEFAULT);
+    assertThat(((Long) entity.getProperty(PROPERTY_DURATION)).intValue())
+        .isEqualTo(DURATION_DEFAULT);
+    assertThat((long) entity.getProperty(PROPERTY_MATCHID)).isEqualTo(MATCHID_DEFAULT);
+    assertThat((long) entity.getProperty(PROPERTY_TIMESTAMP)).isEqualTo(TIMESTAMP_DEFAULT);
```
kmh287, line 103: same question about casts

backend/src/test/java/com/google/sps/MatchDatastoreTest.java
jzwang8: The casts are necessary. I just checked. getProperty() returns type Object.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Yes, we need to store the matchId until the match is returned in SearchMatchServlet

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: If the key is invalid, an IllegalArgumentException is thrown.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: Done.

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: Done.

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: I see, added method foundMatch with javadoc that returns participant with a new match and nulled out availability.

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
tcwanggoogle: sgtm

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,73 @@
...
+    return entity.getKey().getId();
+  }
+
+  /** Return Match from entity, or null if entity is null */
+  private static Match getMatchFromEntity(@Nonnull Entity entity) {
+    String firstParticipantUsername =
+        (String) entity.getProperty(PROPERTY_FIRSTPARTICIPANTUSERNAME);
+    String secondParticipantUsername =
+        (String) entity.getProperty(PROPERTY_SECONDPARTICIPANTUSERNAME);
+    int duration = ((Long) entity.getProperty(PROPERTY_DURATION)).intValue();
```
tcwanggoogle, line 57: Question: Is `Long` the same as 'long`?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,73 @@
...
+    long timestamp = (long) entity.getProperty(PROPERTY_TIMESTAMP);
+
+    return new Match(firstParticipantUsername, secondParticipantUsername, duration, timestamp);
```
tcwanggoogle, line 60: Totally optional: I'd think replacing this whole function with
```
return new Match(
   (String) entity.getProperty(PROPERTY_FIRSTPARTICIPANTUSERNAME),
   (String) entity.getProperty(PROPERTY_SECONDPARTICIPANTUSERNAME),
   ((Long) entity.getProperty(PROPERTY_DURATION)).intValue(),
   (long) entity.getProperty(PROPERTY_TIMESTAMP)
)
```
is totally fine with me. But I guess it's just a personal preference.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,23 +87,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      // Match found, add to match datastore, update participant datastore
+      long matchId = matchDatastore.addMatch(match);
+
+      System.out.println(matchDatastore.getMatchFromId(matchId).toString());
+
+      // Update participant entities with new matchId and null availability
+      Participant currParticipant =
+          participantDatastore.getParticipantFromUsername(match.getSecondParticipantUsername());
+      participantDatastore.addParticipant(currParticipant.foundMatch(matchId));
+      participantDatastore.addParticipant(newParticipant.foundMatch(matchId));
```
tcwanggoogle, line 115: Can we still use a function to encapsulate these?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: Weird datastore quirk, I can't just cast duration to an int because I think ints are stored as long in datastore

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,71 +16,65 @@
...
       // Check if participants are both free for that duration + extra
-      ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
-      ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
-      ZonedDateTime earliestEndTimeAvailable =
-          getEarlier(newEndTimeAvailable, currEndTimeAvailable);
+      long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
+      long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
       boolean compatibleTime =
-          dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
+          (currentTimeMillis + (duration + PADDING_TIME) * MINUTES_TO_MILLIS)
```
kmh287, line 65: TimeUnit.MINUTES.toMillis(duration + PADDING_TIME))

Also I think it would be best to document that PADDING_TIME is in minutes, so maybe `PADDING_TIME_MINS` or `PADDING_TIME_MINUTES`

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
kmh287: There are three different considerations here

int and long are different because long uses a different number of bytes and can supporter larger numbers

`long` and `Long` are different because long is a primitive and `Long` is an object type that "boxes" a long. Try to use `long` wherever possible unless there's a possibility to get `null`

backend/src/test/java/com/google/sps/MatchDatastoreTest.java
kmh287: But can the tests still pass without the casts? Under the hood the assertion will call the .equals() method on the specific concrete Object subclass returned by getProperty, even if the return type is Object.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
kmh287: Got it, thanks

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
kmh287: Up to you how you want to handle that. I could see an argument for throwing or just swallowing the exception here.

backend/src/test/java/com/google/sps/MatchDatastoreTest.java
jzwang8: I get this error when I remove the cast: incompatible types: java.lang.Object cannot be converted to java.lang.Str
ing (same for long)

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: What do you mean by swallowing the exception? This is another instance when removeParticipant will only be called if the participant actually is in the datastore, so I'm not sure what to return. Should I add a try catch, return a boolean, and send out an HTTP response error?

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
jzwang8: The cast to a Long and then intValue() part is clunky, but I get errors when I try to cast directly to an int

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -1,65 +1,86 @@
...
     return timestamp;
   }
+
+  /** Return participant with new matchId and nulled out availability */
+  public Participant foundMatch(long newMatchId) {
+    return new Participant(
+        username,
+        /* startTimeAvailable= */ 0,
+        /* endTimeAvailable= */ 0,
+        /* duration= */ 0,
```
kmh287, line 71: If I understand correctly, the idea is to zero all of these fields out so that the matching algorithm will ignore them due to low values? (Zero minute meeting at midnight on January 1, 1970 😄) 

We can do this in a followup PR but I recommend replacing this with an explicit value keeping track of the matched state. I would recommend using an Enum, even if it ends up just having two values for LOOKING_FOR_MATCH and MATCHED. (As for why to avoid a boolean, my old team used to have an expression: Enums get extended, booleans get deprecated) Then we don't need to depend on implementation details of the matching algorithm to ensure matched users are filtered out. 

WDYT? Fix here or file an issue for the next PR?

jzwang8: Good point, I can fix it here. Enums aren't valid datastore property types, though, so should I just do a boolean?

kmh287: Enums can be converted to and from ints

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,71 +16,65 @@
...
 import com.google.sps.data.Match;
 import com.google.sps.data.Participant;
+import com.google.sps.datastore.ParticipantDatastore;
 import java.time.Clock;
-import java.time.ZonedDateTime;
 import java.util.List;
+import javax.annotation.Nullable;
 
-/** Class used to find a match in a list of Participants with the most recently added Participant */
+/** Class used to find a match for new Participant with unmatched Participants in datastore */
```
TcgVanguardTroll, line 24: Nit: for a new Participant with other unmatched participants in data store

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -74,19 +66,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     }
     JSONObject formDetails = obj.getJSONObject(REQUEST_FORM_DETAILS);
 
-    // Retrieve the timeAvailableUntil input and convert to a UTC ZonedDateTime
-    long timeAvailableUntil = formDetails.getLong(REQUEST_TIME_AVAILABLE_UNTIL);
-    ZoneId zoneId = ZoneId.of("UTC");
-    ZonedDateTime startTimeAvailable = ZonedDateTime.now(Clock.systemUTC());
-    ZonedDateTime endTimeAvailable =
-        ZonedDateTime.ofInstant(Instant.ofEpochMilli(timeAvailableUntil), zoneId);
+    // Retrieve the timeAvailableUntil input and convert to a UTC milliseconds
```
dhruvi618, line 69: `// Retrieve timeAvailableUntil input and calculate startTimeAvailable in terms of milliseconds` ?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,25 +87,48 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-            /* id= */ -1L, username, startTimeAvailable, endTimeAvailable, duration, timestamp);
+            username, startTimeAvailable, endTimeAvailable, duration, /* matchId=*/ 0, timestamp);
 
+    // Get DatastoreService and instiate Match and Participant Datastores
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    MatchDatastore matchDatastore = new MatchDatastore(datastore);
+    ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
+
+    // Check if new participant already in datastore (unmatched, in queue)
+    if (participantDatastore.getParticipantFromUsername() == null) {
```
dhruvi618, line 101: `participantDatastore.getParticipantFromUsername(username) == null`

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -99,25 +87,48 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    participantDatastore.addParticipant(newParticipant);
 
-    // Match found, add to datastore, delete matched participants from datastore
     if (match != null) {
-      addMatchToDatastore(match, datastore);
-      deleteParticipantFromDatastore(match.getSecondParticipant(), datastore);
+      // Match found, add to match datastore, update participant datastore
+      long matchId = matchDatastore.addMatch(match);
+
+      System.out.println(match);
```
dhruvi618, line 117: nit: remove print statement

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
dhruvi618: Oh, just realized that this method call without the param is causing the build to fail

backend/src/test/java/com/google/sps/ParticipantDatastoreTest.java
```diff
@@ -0,0 +1,219 @@
...
+            ENDTIMEAVAILABLE_DEFAULT,
+            DURATION_DEFAULT,
+            MATCHID_DEFAULT,
+            TIMESTAMP_DEFAULT);
+
+    participantDatastore.addParticipant(participant);
+    participantDatastore.removeParticipant(participant.getUsername());
+
+    Participant participantFromUsername = participantDatastore.getParticipantFromUsername(PERSON_A);
+    assertThat(participantFromUsername).isNull();
```
TcgVanguardTroll, line 217: NIt: Do you think you should add test-cases with multiple partcipants and potentially comparing the lengths

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Created MatchStatus enum

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
jzwang8: Yup fixed it, just forgot to mvn test before

backend/src/test/java/com/google/sps/ParticipantDatastoreTest.java
jzwang8: What do you mean? Lengths of what?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -66,10 +74,11 @@ public long getTimestamp() {
...
     return new Participant(
         username,
-        /* startTimeAvailable= */ 0,
-        /* endTimeAvailable= */ 0,
-        /* duration= */ 0,
+        startTimeAvailable,
+        endTimeAvailable,
+        duration,
         newMatchId,
+        matchStatus,
```
kmh287, line 81: update match status

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -90,16 +93,23 @@ private static Participant getParticipantFromEntity(@Nonnull Entity entity) {
         (long) entity.getProperty(PROPERTY_ENDTIMEAVAILABLE),
         ((Long) entity.getProperty(PROPERTY_DURATION)).intValue(),
         (long) entity.getProperty(PROPERTY_MATCHID),
+        MatchStatus.forIntValue(((Long) entity.getProperty(PROPERTY_MATCHID)).intValue()),
```
kmh287, line 96: (long) instead of (Long)?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -26,6 +28,7 @@ public Participant(
       long endTimeAvailable,
       int duration,
       long matchId,
+      MatchStatus matchStatus,
```
kmh287, line 31: A potential cleanup for a future PR could be to make this constructor private and instead use public factory methods to create participants. Some of the code that calls this now has to specify fields that don't make sense for their particular case. For example, creating a new unmatched participant requires the caller to explicitly include the MatchStatus and a zeroed out match ID. Not only does this lead to more LOC, but it also leaks a small detail that 0 is this class' default value for matchId. 

A factory method might look like

```
public Participant createUnmatchedParticipant(String username, long startTimeAvailable, long endTimeAvailable, int duration, long timestamp) {
  return new Participant(username, startTimeAvailable, endTimeAvailable, /* matchId= */ 0, MatchStatus.UNMATCHED, timestamp)
}
```

which requires two fewer arguments because an unmatched participant will always have the unmatched enum value and zeroed match ID. 

Not for this PR, but something to think about.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -16,71 +16,65 @@
...
       // Check if participants are both free for that duration + extra
-      ZonedDateTime newEndTimeAvailable = newParticipant.getEndTimeAvailable();
-      ZonedDateTime currEndTimeAvailable = currParticipant.getEndTimeAvailable();
-      ZonedDateTime earliestEndTimeAvailable =
-          getEarlier(newEndTimeAvailable, currEndTimeAvailable);
+      long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
+      long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
       boolean compatibleTime =
-          dateTime.plusMinutes(duration + PADDING_TIME).isBefore(earliestEndTimeAvailable);
+          (currentTimeMillis + (duration + PADDING_MINUTES) * MINUTES_TO_MILLIS)
```
kmh287, line 65: still could use TimeUnit.MINUTES here to remove the manual conversion to millis

backend/src/main/java/com/google/sps/data/MatchStatus.java
```diff
@@ -0,0 +1,26 @@
+package com.google.sps.data;
+
+/** Represent whether or not a participant is matched */
+public enum MatchStatus {
+  UNMATCHED(0),
+  MATCHED(1);
```
kmh287, line 6: document what each value means (even if it's obvious)

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: Since it's stored as an int, and I can't directly cast to an int, I have to first convert it to a Long and then get the int value. Or I could store it as a long instead of an int so I could just do (long), but I think that'd be more complicated because conversion between ints and enums seems more natural. What do you think?

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Ah interesting. I'll add this as an issue

backend/src/main/java/com/google/sps/data/MatchStatus.java
```diff
@@ -0,0 +1,26 @@
+package com.google.sps.data;
+
+/** Represent whether or not a Participant is matched */
+public enum MatchStatus {
+  UNMATCHED(0), // Not matched yet, waiting for match
+  MATCHED(1); // Matched, can't be matched again yet
```
kmh287, line 6: nit: these should be javadocs above each entry

backend/src/main/java/com/google/sps/datastore/MatchDatastore.java
```diff
@@ -0,0 +1,70 @@
...
+  /** Datastore */
+  private final DatastoreService datastore;
+
+  /** Constructor that takes in DatastoreService */
+  public MatchDatastore(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  /** Return entity of match */
+  private Entity getEntityFromMatch(Match match) {
```
kmh287, line 31: nit: this can be static

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,125 @@
...
+  /** Datastore */
+  private final DatastoreService datastore;
+
+  /** Constructor that takes in DatastoreService */
+  public ParticipantDatastore(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  /** Return entity of participant */
+  private Entity getEntityFromParticipant(Participant participant) {
```
kmh287, line 42: nit: static

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
kmh287: My bad. To "swallow" an exception is to catch it so that it isn't thrown any further up the call stack. So this method could either throw the exception or just catch and log it. The difference is in how the application should behave when this happens. 

If the method throws the servlet will crash and if it catches the exception then it will continue as normal. You might be thinking crashing is always bad but it's also a clear indicator to us the developers that there's an error that we might be able to fix. 

For this specific case, I'm leaning towards catching and logging the exception without throwing anything to the caller. If two servlets want to delete the same entry from the datastore at the same time, there's no reason that one of them should crash. WDYT?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,125 @@
...
+public final class ParticipantDatastore {
+
+  // Datastore Key/Property constants
+  private static final String KIND_PARTICIPANT = "Participant";
+  private static final String PROPERTY_USERNAME = "username";
+  private static final String PROPERTY_STARTTIMEAVAILABLE = "startTimeAvailable";
+  private static final String PROPERTY_ENDTIMEAVAILABLE = "endTimeAvailable";
+  private static final String PROPERTY_DURATION = "duration";
+  private static final String PROPERTY_MATCHID = "matchId";
+  private static final String PROPERTY_MATCHSTATUS = "matchStatus";
```
kmh287, line 30: nit: PROPERTY_MATCH_STATUS

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -0,0 +1,125 @@
...
+  /** Return participant object from datastore participant entity, or null if entity is null */
+  @Nullable
+  private static Participant getParticipantFromEntity(@Nonnull Entity entity) {
+    return new Participant(
+        (String) entity.getProperty(PROPERTY_USERNAME),
+        (long) entity.getProperty(PROPERTY_STARTTIMEAVAILABLE),
+        (long) entity.getProperty(PROPERTY_ENDTIMEAVAILABLE),
+        ((Long) entity.getProperty(PROPERTY_DURATION)).intValue(),
+        (long) entity.getProperty(PROPERTY_MATCHID),
+        MatchStatus.forIntValue(((Long) entity.getProperty(PROPERTY_MATCHID)).intValue()),
```
kmh287, line 95: PROPERTY_MATCH_STATUS

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
kmh287: Ok, I see, it's stored as a Long and so a direct case to int is impossible. Ok to leave as is

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: I see, I changed it so it catches the exception and prints. I'm planning on investigating how to use the logger in a future PR.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -16,42 +16,33 @@
...
-
   // JSON key constants
-  private static final String KEY_MATCHSTATUS = "matchStatus";
-  private static final String KEY_FIRSTPARTICIPANTUSERNAME = "firstParticipantUsername";
-  private static final String KEY_SECONDPARTICIPANTUSERNAME = "secondParticipantUsername";
-  private static final String KEY_DURATION = "duration";
+  private static final String JSON_MATCHSTATUS = "matchStatus";
+  private static final String JSON_FIRSTPARTICIPANTUSERNAME = "firstParticipantUsername";
+  private static final String JSON_SECONDPARTICIPANTUSERNAME = "secondParticipantUsername";
+  private static final String JSON_DURATION = "duration";
```
kmh287, line 41: JSON_MATCH_STATUS, JSON_FIRST_PARTICIPANT_USERNAME etc.

PR #51, opened 2020-07-17
Add form validation algorithm to verify inputs on submission and corresponding tests

dhruvi618:
Files changed: 

- Form.js: Add form validation algorithm to verify inputs on submission in order to prevent incomplete or invalid data to be communicated to the backend.
- Form.test.js: Add tests to ensure accuracy and reliability of `validateFormInputs()` in Form.js
- MatchPreference.js: Update component with a disabled attribute to prevent users from selecting a preference if they have not specified input fields and add helper text to improve user experience and guide user

Snapshot files changed:
- App.test.js.snap: Update snapshot file since the `<App />` component's child component `<MatchPreference />` was updated with helper text
- Form.test.js.snap: Update snapshot file since the `<App />` component's child component `<MatchPreference />` was updated with helper text
- MatchPreference.test.js.snap: Update snapshot since component was updated with helper text and a disabled attribute was added to the radio buttons

### DIFF
--- a/frontend/src/components/Form.js
+++ b/frontend/src/components/Form.js
@@ -89,11 +89,14 @@ export default function Form(props) {
   const [matchPreference, setMatchPreference] = React.useState('none');
 
   /**
-   * @return {Boolean} true/false based on completion status of personal fields
+   * Method that controls the disabled attribute of MatchPreference radio group
+   * based on the completion status of the personal preference fields
+   * @return {Boolean} true/false where true means the MatchPreference options
+   * will be disabled and false means that the options will be enabled.
    */
-  function checkPersonalPreferenceFields() {
+  function toggleMatchPreferenceOptions() {
     // TO-DO(#65): Add interests to this validation
-    return role === '' || productArea === '';
+    return role === '' && productArea === '';
   }
 
   /** Gather user inputs from form and send POST request to backend
@@ -169,7 +172,7 @@ export default function Form(props) {
         <div className={classes.padding}>
           <MatchPreference
             onChange={(value) => setMatchPreference(value)}
-            checkPersonalPreferenceFields={checkPersonalPreferenceFields()}
+            checkPersonalPreferenceFields={toggleMatchPreferenceOptions()}
           />
         </div>
       </div>
--- a/frontend/src/components/Form.js
+++ b/frontend/src/components/Form.js
@@ -47,19 +47,25 @@ const useStyles = makeStyles((theme) => ({
  * @param {String} role
  * @param {String} productArea
  * @param {Number} duration
- * @param {Number} timeAvailableUntilMilliseconds
- * @param {Number} currentTimeInMilliseconds
+ * @param {Date} timeAvailableUntil
+ * @param {Number} currentTimeMilliseconds
  */
 export function validateFormInputs(role, productArea, duration,
-    timeAvailableUntilMilliseconds, currentTimeInMilliseconds) {
-  const durationInMilliseconds = duration * 60000;
-  if (productArea === '' || role === '') {
-    alert('Please select options for all form fields to submit!');
+    timeAvailableUntil, currentTimeMilliseconds) {
+  if (!(timeAvailableUntil instanceof Date) ||
+    isNaN(timeAvailableUntil.getTime())) {
+    alert('Please select a valid date.');
     return false;
-  } else if (currentTimeInMilliseconds + durationInMilliseconds >=
-      timeAvailableUntilMilliseconds) {
+  }
+  const timeAvailableUntilMilliseconds = timeAvailableUntil.getTime();
+  const durationMilliseconds = duration * 60000;
+  if (currentTimeMilliseconds + durationMilliseconds >=
+    timeAvailableUntilMilliseconds) {
     // Check if a meeting is possible with the provided inputs
     alert('Please select an larger time availability window');
+    return false;
+  } else if (productArea === '' || role === '') {
+    alert('Please select options for all form fields to submit!');
     return false;
   } else {
     return true;
@@ -88,7 +94,7 @@ export default function Form() {
   function handleFormSubmission(event) {
     const currentTimeInMilliseconds = new Date().getTime();
     if (validateFormInputs(role, productArea, duration,
-        timeAvailableUntil.getTime(), currentTimeInMilliseconds)) {
+        timeAvailableUntil, currentTimeInMilliseconds)) {
       // Override browser's default behvaior to execute POST request
       event.preventDefault();
 
--- a/frontend/src/components/Form.js
+++ b/frontend/src/components/Form.js
@@ -47,13 +47,13 @@ const useStyles = makeStyles((theme) => ({
  * @param {String} role
  * @param {String} productArea
  * @param {Number} duration
- * @param {Date} timeAvailableUntil
+ * @param {Number} timeAvailableUntilMilliseconds
  * @param {Number} currentTimeMilliseconds
  */
 export function validateFormInputs(role, productArea, duration,
     timeAvailableUntilMilliseconds, currentTimeMilliseconds) {
   const durationMilliseconds = duration * 60000;
-  
+
   if (isNaN(timeAvailableUntilMilliseconds)) {
     alert('Please select a valid date.');
     return false;
--- a/frontend/src/components/Form.js
+++ b/frontend/src/components/Form.js
@@ -42,6 +42,31 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 /**
+ * Validates from inputs on submission and alerts user on error
+ * @return {Boolean}  true or false based on validity of inputs
+ * @param {String} role
+ * @param {String} productArea
+ * @param {Number} duration
+ * @param {Number} timeAvailableUntilMilliseconds
+ * @param {Number} currentTimeInMilliseconds
+ */
+export function validateFormInputs(role, productArea, duration,
+    timeAvailableUntilMilliseconds, currentTimeInMilliseconds) {
+  const durationInMilliseconds = duration * 60000;
+  if (productArea === '' || role === '') {
+    alert('Please select options for all form fields to submit!');
+    return false;
+  } else if (currentTimeInMilliseconds + durationInMilliseconds >=
+      timeAvailableUntilMilliseconds) {
+    // Check if a meeting is possible with the provided inputs
+    alert('Please select an larger time availability window');
+    return false;
+  } else {
+    return true;
+  }
+}
+
+/**
  * Create form component with time and match preference inputs
  * @return {Form} Form component
  */
@@ -57,27 +82,13 @@ export default function Form() {
   const [savePreference, setSavePreference] = React.useState(true);
   const [matchPreference, setMatchPreference] = React.useState('none');
 
-  function validateFormInputs() {
-    const currentTimeInMilliseconds = new Date().getTime();
-    const durationInMilliseconds = duration * 60000;
-    if (productArea === '' || role === '') {
-      alert("Please select options for all form fields to submit!");
-      return false;
-    } else if (currentTimeInMilliseconds + durationInMilliseconds
-        >= timeAvailableUntil.getTime()) {
-      // Check if a meeting is possible with the provided inputs
-      alert("Please select an larger time availability window");
-      return false;
-    } else {
-      return true;
-    }
-  }
-
   /** Gather user inputs from form and send POST request to backend
     * @param {Event} event
    */
   function handleFormSubmission(event) {
-    if (validateFormInputs()) {
+    const currentTimeInMilliseconds = new Date().getTime();
+    if (validateFormInputs(role, productArea, duration,
+        timeAvailableUntil.getTime(), currentTimeInMilliseconds)) {
       // Override browser's default behvaior to execute POST request
       event.preventDefault();
 
@@ -95,7 +106,7 @@ export default function Form() {
       axios.post('/api/v1/add-participant', {formDetails})
           .then((response) => {
             if (response.data != null) {
-              // TODO(#33): change alert to a redirection to the loading page view
+              // TODO(#33): change alert to a redirection to loading page view
               alert('Successful');
             }
           });
--- a/frontend/src/components/Form.js
+++ b/frontend/src/components/Form.js
@@ -57,20 +57,27 @@ export default function Form() {
   const [savePreference, setSavePreference] = React.useState(true);
   const [matchPreference, setMatchPreference] = React.useState('none');
 
+  function validateFormInputs() {
+    const currentTimeInMilliseconds = new Date().getTime();
+    const durationInMilliseconds = duration * 60000;
+    if (productArea === '' || role === '') {
+      alert("Please select options for all form fields to submit!");
+      return false;
+    } else if (currentTimeInMilliseconds + durationInMilliseconds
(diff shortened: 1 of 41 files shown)


frontend/src/components/Form.js
```diff
@@ -41,11 +41,41 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+/**
+ * Validates from inputs on submission and alerts user on error
+ * @return {Boolean} true or false based on validity of inputs
+ * @param {String} role
+ * @param {String} productArea
+ * @param {Number} duration
+ * @param {Number} timeAvailableUntilMilliseconds
```
jzwang8, line 50: change to endTimeAvailable eventually?

frontend/src/components/Form.js
```diff
@@ -41,11 +41,41 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+/**
+ * Validates from inputs on submission and alerts user on error
+ * @return {Boolean} true or false based on validity of inputs
```
kmh287, line 46: Should this go after all the @param comments?

frontend/src/components/Form.js
```diff
@@ -41,11 +41,41 @@ const useStyles = makeStyles((theme) => ({
...
+ * @param {String} role
+ * @param {String} productArea
+ * @param {Number} duration
+ * @param {Number} timeAvailableUntilMilliseconds
+ * @param {Number} currentTimeMilliseconds
+ */
+export function validateFormInputs(role, productArea, duration,
+    timeAvailableUntilMilliseconds, currentTimeMilliseconds) {
```
kmh287, line 54: nit: if all the parameters can't fit on one line, put each one on its own line

frontend/src/components/Form.js
```diff
@@ -41,11 +41,41 @@ const useStyles = makeStyles((theme) => ({
...
+    timeAvailableUntilMilliseconds) {
+    // Check if a meeting is possible with the provided inputs
+    alert('Please select an larger time availability window');
+    return false;
+  } else if (productArea === '' || role === '') {
+    alert('Please select options for all form fields to submit!');
+    return false;
+  } else {
+    return true;
+  }
```
kmh287, line 70: tiny nit:

no need to include an else here since all the if-branches return i.e. this can just be `return true` outside of an else {}

frontend/src/components/Form.js
```diff
@@ -61,29 +91,33 @@ export default function Form() {
     * @param {Event} event
    */
   function handleFormSubmission(event) {
-    // TODO(#34): Validate input values (before allowing to submit)
+    const currentTimeInMilliseconds = new Date().getTime();
+    if (validateFormInputs(role, productArea, duration,
+        timeAvailableUntil.getTime(), currentTimeInMilliseconds)) {
```
kmh287, line 96: If this can't fit on a single line, you can do 

```
(validateFormInputs(
    role, productArea, duration, timeAvailableUntil.getTime(), currentTimeInMilliseconds))
```

and if that still can't fit on a single line then 

```
(validateFormInputs(
    role, 
    productArea, 
    duration,
    timeAvailableUntil.getTime(),
    currentTimeInMilliseconds))
```

frontend/src/components/Form.js
```diff
@@ -61,29 +91,33 @@ export default function Form() {
...
 
-    // Send form details to AddParticipantServlet and alert user on success
-    axios.post('/api/v1/add-participant', {formDetails})
-        .then((response) => {
-          if (response.data != null) {
-            // TODO(#33): change alert to a redirection to the loading page view
-            alert('Successful');
-          }
-        });
+      // TO-DO: call getMatch() to initiate GET request to search-match servlet
```
kmh287, line 119: add an issue ID

frontend/src/components/Form.test.js
```diff
@@ -45,3 +46,67 @@ describe('Form', () => {
     );
   });
 });
+
+describe('Form Validation', () => {
+  it('should return true', () => {
```
kmh287, line 51: should be true when...?

frontend/src/components/Form.test.js
```diff
@@ -45,3 +46,67 @@ describe('Form', () => {
...
+    // Wed Jul 15 2020 20:56:45
+    const timeAvailableUntilMilliseconds = 1594846605591;
+    const duration = 15;
+    // Wed Jul 15 2020 08:30:00
+    const currentTimeMilliseconds = 1594801800000;
+    expect(validateFormInputs(role, productArea, duration,
+        timeAvailableUntilMilliseconds, currentTimeMilliseconds)).toBe(true);
+  });
+
+  it('should return false since Role is not specified', () => {
```
kmh287, line 63: tiny nit, here and below s/since/when

frontend/src/components/Form.js
dhruvi618: Yes, I changed this in the following PR #64 . Thanks!

frontend/src/components/Form.js
dhruvi618: Added, thanks!

frontend/src/components/Form.js
dhruvi618: Used the second version and adjusted it to fix any associated linter errors

frontend/src/components/Form.js
dhruvi618: Good point, thanks for the suggestion!

frontend/src/components/Form.js
dhruvi618: Done, thanks!

frontend/src/components/Form.js
```diff
@@ -41,11 +41,45 @@ const useStyles = makeStyles((theme) => ({
...
+  if (isNaN(timeAvailableUntilMilliseconds)) {
+    alert('Please select a valid date.');
+    return false;
+  } else if (currentTimeMilliseconds + durationMilliseconds >=
+    timeAvailableUntilMilliseconds) {
+    // Check if a meeting is possible with the provided inputs
+    alert('Please select an larger time availability window');
+    return false;
+  } else if (productArea === '' || role === '') {
+    alert('Please select options for all form fields to submit!');
```
tcwanggoogle, line 71: For these two fields, once we have "no preference" match checkbox, we should allow them to be empty, right?

frontend/src/components/Form.js
```diff
@@ -41,11 +41,45 @@ const useStyles = makeStyles((theme) => ({
...
+  return true;
+}
+
 /**
  * Create form component with time and match preference inputs
+ * @param {Object} props
  * @return {Form} Form component
  */
-export default function Form() {
+export default function Form(props) {
```
tcwanggoogle, line 82: Any reason why we need to add `props` to the arguments?

frontend/src/components/Form.js
```diff
@@ -61,29 +95,37 @@ export default function Form() {
...
    */
   function handleFormSubmission(event) {
-    // TODO(#34): Validate input values (before allowing to submit)
+    const currentTimeInMilliseconds = new Date().getTime();
+    if (validateFormInputs(
+        role,
+        productArea,
+        duration,
+        timeAvailableUntil.getTime(),
+        currentTimeInMilliseconds)) {
```
tcwanggoogle, line 104: It looks like all the following code in the function is for this if? If so, please flip the condition and return early if false. This can prevent my question of "where the if ends".

frontend/src/components/Form.test.js
dhruvi618: Changed to "should return ____ when..." format

frontend/src/components/Form.js
dhruvi618: This is used for communication purposes so that the child component can communicate with the parent component in the future.

frontend/src/components/Form.js
dhruvi618: Thanks for the suggestions, done!

frontend/src/components/Form.js
TcgVanguardTroll: + 1

dhruvi618: I think we were discussing that we want to require these fields regardless so that the database doesn't have empty entries, but I think is a great suggestion. I will follow up with the rest of the pod on thoughts and update. Thanks

frontend/src/components/Form.js
```diff
@@ -41,11 +41,45 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+/**
+ * Validates from inputs on submission and alerts user on error
+ * @param {String} role
+ * @param {String} productArea
+ * @param {Number} duration
```
kmh287, line 48: document the time unit here. From the conversion below it looks like minutes?

frontend/src/components/Form.test.js
```diff
@@ -45,3 +46,68 @@ describe('Form', () => {
...
+  it('should return true when all fields are provided and time is compatible',
+      () => {
+        const role = 'Intern';
+        const productArea = 'Platforms and Ecosystems';
+        // Wed Jul 15 2020 20:56:45
+        const endTimeAvailableMilliseconds = 1594846605591;
+        const duration = 15;
+        // Wed Jul 15 2020 08:30:00
+        const currentTimeMilliseconds = 1594801800000;
+        expect(validateFormInputs(role, productArea, duration,
```
kmh287, line 60:
```suggestion
        const currentTimeMilliseconds = 1594801800000;
        
        expect(validateFormInputs(role, productArea, duration,
```

newline between these two separate sections of the test. Here and below

frontend/src/components/Form.js
kmh287: Also wondering about this. Maybe the default option can be something like "Prefer not to say" to make it clear to the user that declining to pick something will have an impact on matching.

dhruvi618: After discussing we came to this conclusion: We need to store the Role and PA  in case the other person they would match has a preference. For example, if Person A specifies 'similar' and Person B specifies 'any' and both A and B are indeed similar a match between A and B could occur. But, if B does not specify their Role and PA, then A and B would not be able to match because their is no point of comparison between the two since B did not supply any inputs.

frontend/src/components/Form.js
dhruvi618: Added!

frontend/src/components/Form.test.js
dhruvi618: Made this change, and I have a follow up question to potentially clean it up more: Based on the arrange-act-assert test organization method, would this be cleaner if I made the method call and saved the value to a variable in a separate line above the assertion step?

frontend/src/components/Form.js
kmh287: Thanks for the explanation Dhruvi. One possible alternative could be to lock the match preference to ANY (along with an explanation) until the user specifies at least one field on which they can be matched. That way a user can still participate even if they don't want to specify role and PA. WDYT?

frontend/src/components/Form.test.js
dhruvi618: Something like:

// Arrange step with all values
`const role = 'Intern' `
...

// Act
`const inputValidation = validateFormInputs(role, productArea, duration, endTimeAvailableMilliseconds, currentTimeMilliseconds)`

// Assert 
`expect(inputValidation).toBe(true);`

kmh287: It couldn't hurt but it's not required if the expected value can be calculated concisely.

```
it('should add numbers', () => {
  expect(adder.add(2,2)).toBe(4)
})
```

is a perfectly valid test.

dhruvi618: Got it, thanks for the clarification!

frontend/src/components/Form.js
dhruvi618: I like this idea! It may cause backend implementation to change, but I think it is a good way to get a user to participate without specifying personal details as you mentioned. I will work on this change and update.

tcwanggoogle: Thanks for the discussion. If it needs backend implementation to change, we should use a follow up PR to change frontend and backend together to support empty role/PA field. IMO, a TODO here would be enough for this PR.

dhruvi618:
@tcwanggoogle Sorry, I just saw your comment about adding these changes into a separate PR after I made the corresponding changes to the frontend. I will add an issue to implement these in the backend as well so we can keep track of this.

frontend/src/components/Form.js
```diff
@@ -91,6 +91,14 @@ export default function Form(props) {
   const [savePreference, setSavePreference] = React.useState(true);
   const [matchPreference, setMatchPreference] = React.useState('none');
 
+  /**
+   * @return {Boolean} true/false based on completion status of personal fields
+   */
+  function checkPersonalPreferenceFields() {
+    // TO-DO(#65): Add interests to this validation
+    return role === '' || productArea === '';
+  }
```
kmh287, line 97: So if both role and product area are empty then the fields are complete?

dhruvi618: Yes. So you no longer have to submit a role and product area (and later interests as well) to submit the match form. In the case that you do not submit role and PA, the "similar" and "different" match preference radio buttons are disabled:
Screenshot 2020-07-22 at 11 17 09 AM

As soon as a role and PA are selected, you are able to submit a preference again!

jzwang8: Also, is the UI change we discussed earlier about dividing the "choose your match preferences" section into two (Describe yoruself, Who do you want to meet?) coming in a later PR?

kmh287: I guess it's not clear to me when this method should return true. The name suggests it should return true if both fields are filled out but instead it returns true if both fields are empty.

dhruvi618: > Also, is the UI change we discussed earlier about dividing the "choose your match preferences" section into two (Describe yoruself, Who do you want to meet?) coming in a later PR?

Yes, this will come in a later PR

frontend/src/components/Form.js
```diff
@@ -89,11 +89,14 @@ export default function Form(props) {
...
 
   /**
-   * @return {Boolean} true/false based on completion status of personal fields
+   * Method that controls the disabled attribute of MatchPreference radio group
+   * based on the completion status of the personal preference fields
+   * @return {Boolean} true/false where true means the MatchPreference options
+   * will be disabled and false means that the options will be enabled.
    */
-  function checkPersonalPreferenceFields() {
+  function toggleMatchPreferenceOptions() {
```
kmh287, line 97: A method that returns true when something should be turned off and false when it should be turned on seems counterintuitive. How about 

```
function shouldEnableMatchPReferences() {
  return role != '' || productArea != ''
}
```

dhruvi618: I understand this may be counterintuitive, however, the match preferences need to be disabled by default since the page will load with the role and PA empty and once either role or PA or both are filled out the preferences should be enabled again. Using this suggestion, the opposite would occur since role and product area will be empty to start off and so the method will return false which will cause the disabled attribute to be false. To make it clearer, I can change the one line method into a if-else statement and make the return values more intuitive, would that work?

kmh287: My concern is more that the return value of the method isn't intuitive from the method name. The use is here

```
          disabled={props.checkPersonalPreferenceFields}
```

so maybe the method could be called `shouldDisableMatchPreferenceField`. WDYT?

dhruvi618: Retained the original logic, changed the method name.

PR #55, opened 2020-07-20
Remove expired participants

jzwang8:
File changed:
- SearchMatchServlet: remove participant from participant datastore if currently unmatched and time has expired, send "expired" JSON response

### DIFF
--- a/backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
+++ b/backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
@@ -24,6 +24,7 @@ import com.google.sps.data.Participant;
 import com.google.sps.datastore.MatchDatastore;
 import com.google.sps.datastore.ParticipantDatastore;
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -34,6 +35,9 @@ import org.json.simple.JSONObject;
 @WebServlet("/api/v1/search-match")
 public class SearchMatchServlet extends HttpServlet {
 
+  /** Extra padding time in minutes to ensure large enough meeting time block */
+  private static final int PADDING_MINUTES = 10;
+
   // JSON key constants
   private static final String JSON_MATCH_STATUS = "matchStatus";
   private static final String JSON_FIRST_PARTICIPANT_USERNAME = "firstParticipantUsername";
@@ -43,6 +47,7 @@ public class SearchMatchServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     System.out.println("Request received");
+
     // Get participant username
     UserService userService = UserServiceFactory.getUserService();
     String email = userService.getCurrentUser().getEmail();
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -92,14 +97,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     participantDatastore.removeParticipant(match.getFirstParticipantUsername());
     participantDatastore.removeParticipant(match.getSecondParticipantUsername());
 
+    sendMatchResponse(response, match);
+  }
+
+  /** Return true if expired and should be removed, false if still valid */
+  private boolean checkExpired(Participant participant) {
```
kmh287, line 104: isExpired, since it returns a boolean?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -92,14 +97,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     participantDatastore.removeParticipant(match.getFirstParticipantUsername());
     participantDatastore.removeParticipant(match.getSecondParticipantUsername());
 
+    sendMatchResponse(response, match);
+  }
+
+  /** Return true if expired and should be removed, false if still valid */
```
kmh287, line 103: Can you expand on this to explain when a participant is expired?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -92,14 +97,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  private boolean checkExpired(Participant participant) {
+    long currentTimeMillis = System.currentTimeMillis();
+    return (participant.getEndTimeAvailable()
```
kmh287, line 106: This took me a bit to parse so a comment might be helpful to any future readers. To make sure I understand correctly, if I'm available until 2pm and want a meeting duration of 30 minutes, I'll be removed as expired 40 minutes before 2pm at 1:20pm (30 minutes + 10 minute padding)?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -92,14 +97,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    JSONObject expired = new JSONObject();
+    expired.put(JSON_MATCH_STATUS, "expired");
+
+    // Send the JSON back as the response
+    response.setContentType("application/json");
+    try {
+      response.getWriter().println(expired.toString());
+    } catch (IOException e) {
+      System.out.println("Failed to return expired participant response.");
+    }
```
kmh287, line 122: Before this refactor doGet would throw these IOExceptions but now they'll be swallowed in these methods. Was that a deliberate change or can we change the methods signatures to include `throws IOException` and remove the catches?

dhruvi618: +1. Maybe we can throw the exceptions with the error messages you are outputting

kmh287:
As we add more and more functionality to this class, its lack of testing becomes a larger issue. This change LGTM but this class is due for some refactoring and testing. I'd recommend a followup PR just focused on refactoring and adding tests.

PR #62, opened 2020-07-20
Add MVP fields to matching algorithm

jzwang8:
Files changed:
- Participant/ParticipantDatastore/ParticipantDatastoreTest: add personal fields and test
- User/UserDatastore/UserDatastoreTest: add personal fields and test
- FindMatchQuery/FindMatchQueryTest: include personal fields/match preference logic to match query and test
- AddParticipantServlet: use form input fields, add user if opted to save preference
- MatchPreference: enum for preference of being matched with similar, any, or different Googler

### DIFF
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -14,6 +14,7 @@
 
 package com.google.sps;
 
+import com.google.common.primitives.Booleans;
 import com.google.sps.data.Match;
 import com.google.sps.data.MatchPreference;
 import com.google.sps.data.Participant;
@@ -53,9 +54,6 @@ public final class FindMatchQuery {
     List<Participant> sameDurationParticipants =
         participantDatastore.getParticipantsWithDuration(duration);
 
-    // Set reference date time using clock
-    long currentTimeMillis = clock.millis();
-
     long newEndTimeAvailable = newParticipant.getEndTimeAvailable();
     MatchPreference newMatchPreference = newParticipant.getMatchPreference();
     String newRole = newParticipant.getRole();
@@ -65,8 +63,7 @@ public final class FindMatchQuery {
     for (Participant currParticipant : sameDurationParticipants) {
       // Check if participants are both free for that duration + extra
       long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
-      if (!isCompatibleTime(
-          currentTimeMillis, duration, newEndTimeAvailable, currEndTimeAvailable)) {
+      if (!isCompatibleTime(duration, newEndTimeAvailable, currEndTimeAvailable)) {
         System.out.println("not compatible time");
         continue;
       }
@@ -92,7 +89,7 @@ public final class FindMatchQuery {
       }
       System.out.println("found a match");
       return new Match(
-          newParticipant.getUsername(), currParticipant.getUsername(), duration, currentTimeMillis);
+          newParticipant.getUsername(), currParticipant.getUsername(), duration, clock.millis());
     }
     // No inital match found
     return null;
@@ -100,9 +97,9 @@ public final class FindMatchQuery {
 
   /** Return true if compatible endTimeAvailable considering duration and padding */
   private boolean isCompatibleTime(
-      long currentTimeMillis, int duration, long newEndTimeAvailable, long currEndTimeAvailable) {
+      int duration, long newEndTimeAvailable, long currEndTimeAvailable) {
     long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
-    return (currentTimeMillis + TimeUnit.MINUTES.toMillis(duration + PADDING_MINUTES))
+    return (clock.millis() + TimeUnit.MINUTES.toMillis(duration + PADDING_MINUTES))
         < earliestEndTimeAvailable;
   }
 
@@ -125,13 +122,9 @@ public final class FindMatchQuery {
 
   /** Return the number of fields that both participants have in common */
   private int getNumSameFields(Participant newParticipant, Participant currParticipant) {
-    int numSameFields = 0;
-    if (newParticipant.getRole().equals(currParticipant.getRole())) {
-      numSameFields++;
-    }
-    if (newParticipant.getProductArea().equals(currParticipant.getProductArea())) {
-      numSameFields++;
-    }
-    return numSameFields;
+    boolean sameRole = newParticipant.getRole().equals(currParticipant.getRole());
+    boolean sameProductArea =
+        newParticipant.getProductArea().equals(currParticipant.getProductArea());
+    return Booleans.countTrue(sameRole, sameProductArea);
   }
 }
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -54,39 +54,36 @@ public final class FindMatchQuery {
     List<Participant> sameDurationParticipants =
         participantDatastore.getParticipantsWithDuration(duration);
 
-    long newEndTimeAvailable = newParticipant.getEndTimeAvailable();
-    MatchPreference newMatchPreference = newParticipant.getMatchPreference();
-    String newRole = newParticipant.getRole();
-    String newProductArea = newParticipant.getProductArea();
+    long firstEndTimeAvailable = newParticipant.getEndTimeAvailable();
+    MatchPreference firstMatchPreference = newParticipant.getMatchPreference();
+    String firstRole = newParticipant.getRole();
+    String firstProductArea = newParticipant.getProductArea();
 
     // Compare new participant preferences with other participants to find match
     for (Participant currParticipant : sameDurationParticipants) {
       // Check if participants are both free for that duration + extra
-      long currEndTimeAvailable = currParticipant.getEndTimeAvailable();
-      if (!isCompatibleTime(duration, newEndTimeAvailable, currEndTimeAvailable)) {
+      long secondEndTimeAvailable = currParticipant.getEndTimeAvailable();
+      if (!isCompatibleTime(duration, firstEndTimeAvailable, secondEndTimeAvailable)) {
         System.out.println("not compatible time");
         continue;
       }
 
       // Check match preference compatibility and get combined preference if compatible
-      MatchPreference currMatchPreference = currParticipant.getMatchPreference();
-      if (!isCompatibleMatchPreference(newMatchPreference, currMatchPreference)) {
+      MatchPreference secondMatchPreference = currParticipant.getMatchPreference();
+      MatchPreference combinedMatchPreference =
+          getCombinedMatchPreference(firstMatchPreference, secondMatchPreference);
+      if (combinedMatchPreference == null) {
         System.out.println("not compatible match pref");
         continue;
       }
-      MatchPreference combinedMatchPreference =
-          getCombinedMatchPreference(newMatchPreference, currMatchPreference);
 
-      if (combinedMatchPreference != MatchPreference.ANY) {
-        int numSameFields = getNumSameFields(newParticipant, currParticipant);
-        System.out.println(numSameFields);
-        if (combinedMatchPreference == MatchPreference.SIMILAR && numSameFields < MIN_SAME_FIELDS) {
-          continue;
-        } else if (combinedMatchPreference == MatchPreference.DIFFERENT
-            && numSameFields >= MIN_SAME_FIELDS) {
-          continue;
-        }
+      // Check if combined match preference is satisfied depending on number of same fields
+      if (!isCombinedMatchPreferenceSatisfied(
+          combinedMatchPreference, newParticipant, currParticipant)) {
+        continue;
       }
+
+      // Found a match
       System.out.println("found a match");
       return new Match(
           newParticipant.getUsername(), currParticipant.getUsername(), duration, clock.millis());
@@ -97,18 +94,28 @@ public final class FindMatchQuery {
 
   /** Return true if compatible endTimeAvailable considering duration and padding */
   private boolean isCompatibleTime(
-      int duration, long newEndTimeAvailable, long currEndTimeAvailable) {
-    long earliestEndTimeAvailable = Math.min(newEndTimeAvailable, currEndTimeAvailable);
+      int duration, long firstEndTimeAvailable, long secondEndTimeAvailable) {
+    long earliestEndTimeAvailable = Math.min(firstEndTimeAvailable, secondEndTimeAvailable);
     return (clock.millis() + TimeUnit.MINUTES.toMillis(duration + PADDING_MINUTES))
         < earliestEndTimeAvailable;
   }
 
-  /** Return true if compatible match preferences (same preference, one or both are any) */
-  private boolean isCompatibleMatchPreference(
-      MatchPreference newMatchPreference, MatchPreference currMatchPreference) {
-    return (newMatchPreference == currMatchPreference)
-        || (newMatchPreference == MatchPreference.ANY)
-        || (currMatchPreference == MatchPreference.ANY);
+  /**
+   * @return combined MatchPreference similar = both are similar OR one is similar and one is any;
+   *     any = both are any; different = both are different = one is different and one is any. or
+   *     null if match preferences are not compatible
+   */
+  private MatchPreference getCompatibleMatchPreference(
+      MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
+    if ((firstMatchPreference == secondMatchPreference)
+        || (firstMatchPreference == MatchPreference.ANY)
+        || (secondMatchPreference == MatchPreference.ANY)) {
+      return (firstMatchPreference == MatchPreference.ANY)
+          ? secondMatchPreference
+          : firstMatchPreference;
+    } else {
+      return null;
+    }
   }
 
   /**
@@ -116,15 +123,38 @@ public final class FindMatchQuery {
    *     both are any different: both are different, one is different and one is any
    */
   private MatchPreference getCombinedMatchPreference(
-      MatchPreference newMatchPreference, MatchPreference currMatchPreference) {
-    return (newMatchPreference == MatchPreference.ANY) ? currMatchPreference : newMatchPreference;
+      MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
+    return (firstMatchPreference == MatchPreference.ANY)
+        ? secondMatchPreference
+        : firstMatchPreference;
   }
 
-  /** Return the number of fields that both participants have in common */
-  private int getNumSameFields(Participant newParticipant, Participant currParticipant) {
-    boolean sameRole = newParticipant.getRole().equals(currParticipant.getRole());
-    boolean sameProductArea =
-        newParticipant.getProductArea().equals(currParticipant.getProductArea());
-    return Booleans.countTrue(sameRole, sameProductArea);
+  /**
+   * @return true if the participants are a match based on their combinedMatchPreference and the
+   *     number of fields that are the same, false if not a match
+   */
+  private boolean isCombinedMatchPreferenceSatisfied(
+      MatchPreference combinedMatchPreference,
+      Participant newParticipant,
+      Participant currParticipant) {
+    if (combinedMatchPreference != MatchPreference.ANY) {
+      // Count number of same fields between the two participants
+      boolean sameRole = newParticipant.getRole().equals(currParticipant.getRole());
+      boolean sameProductArea =
+          newParticipant.getProductArea().equals(currParticipant.getProductArea());
+      int numSameFields = Booleans.countTrue(sameRole, sameProductArea);
+      System.out.println("number of same fields: " + numSameFields);
+
+      // If match preference = similar, numSameFields must be at least MIN_SAME_FIELDS to satisfy
+      // preference
+      // If match preference = differe, numSameFields must be less than MIN_SAME_FIELDS to satisfy
+      // preference
+      if ((combinedMatchPreference == MatchPreference.SIMILAR && numSameFields < MIN_SAME_FIELDS)
+          || (combinedMatchPreference == MatchPreference.DIFFERENT
+              && numSameFields >= MIN_SAME_FIELDS)) {
+        return false;
+      }
+    }
+    return true;
   }
 }
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -96,15 +96,16 @@ public final class FindMatchQuery {
    */
   private MatchPreference getCombinedMatchPreference(
       MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
-    if ((firstMatchPreference != secondMatchPreference)
-        && (firstMatchPreference != MatchPreference.ANY)
-        && (secondMatchPreference != MatchPreference.ANY)) {
-      // One user is SIMILAR, other is DIFFERENT. Incompatible.
-      return null;
+    if (firstMatchPreference == secondMatchPreference) {
+      return firstMatchPreference;
     }
-    return (firstMatchPreference == MatchPreference.ANY)
-        ? secondMatchPreference
-        : firstMatchPreference;
+    if (firstMatchPreference == MatchPreference.ANY) {
+      return secondMatchPreference;
+    }
+    if (secondMatchPreference == MatchPreference.ANY) {
+      return firstMatchPreference;
+    }
+    return null;
   }
 
   /**
--- a/backend/src/main/java/com/google/sps/FindMatchQuery.java
+++ b/backend/src/main/java/com/google/sps/FindMatchQuery.java
@@ -136,7 +136,7 @@ public final class FindMatchQuery {
 
       // If match preference = similar, numSameFields must be at least MIN_SAME_FIELDS to satisfy
       // preference
-      // If match preference = differe, numSameFields must be less than MIN_SAME_FIELDS to satisfy
+      // If match preference = different, numSameFields must be less than MIN_SAME_FIELDS to satisfy
       // preference
       if ((combinedMatchPreference == MatchPreference.SIMILAR && numSameFields < MIN_SAME_FIELDS)
           || (combinedMatchPreference == MatchPreference.DIFFERENT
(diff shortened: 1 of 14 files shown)


backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -54,26 +57,81 @@ public Match findMatch(Participant newParticipant) {
...
-
-      if (compatibleTime) {
-        // Create and return match
-        return new Match(
-            newParticipant.getUsername(),
-            currParticipant.getUsername(),
-            duration,
-            currentTimeMillis);
+      if (!isCompatibleTime(
+          currentTimeMillis, duration, newEndTimeAvailable, currEndTimeAvailable)) {
```
kmh287, line 69: can isCompatibleTime call clock.millis() internally?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -54,26 +57,81 @@ public Match findMatch(Participant newParticipant) {
...
+        < earliestEndTimeAvailable;
+  }
+
+  /** Return true if compatible match preferences (same preference, one or both are any) */
+  private boolean isCompatibleMatchPreference(
+      MatchPreference newMatchPreference, MatchPreference currMatchPreference) {
+    return (newMatchPreference == currMatchPreference)
+        || (newMatchPreference == MatchPreference.ANY)
+        || (currMatchPreference == MatchPreference.ANY);
+  }
```
kmh287, line 115: This may produce some unintended results. For example if I ask for a match who is dissimilar to me, and someone very similar to me asks for any match, this logic would allow us to be matched since the other person's match preference is "ANY".

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -54,26 +57,81 @@ public Match findMatch(Participant newParticipant) {
...
+
+  /** Return the number of fields that both participants have in common */
+  private int getNumSameFields(Participant newParticipant, Participant currParticipant) {
+    int numSameFields = 0;
+    if (newParticipant.getRole().equals(currParticipant.getRole())) {
+      numSameFields++;
+    }
+    if (newParticipant.getProductArea().equals(currParticipant.getProductArea())) {
+      numSameFields++;
+    }
```
kmh287, line 134: https://guava.dev/releases/19.0/api/docs/com/google/common/primitives/Booleans.html#countTrue(boolean...)

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -14,6 +28,12 @@
   private final long endTimeAvailable;
   /** How long user wants to chat */
   private final int duration;
+  /** Role at Google */
+  private final String role;
+  /** Product area at Google */
+  private final String productArea;
```
kmh287, line 34: These fields seem more appropriate for the User object than Participant. Could they move to User instead? FindMatchQuery could load the corresponding User models for participants as well.

backend/src/main/java/com/google/sps/data/User.java
```diff
@@ -1,31 +1,76 @@
...
 
 /** A user with saved preferences. */
 public final class User {
 
-  /** Datastore ID */
-  private final long id;
   /** Google username (ldap) */
   private final String username;
+  /** How long user wants to chat */
+  private final int duration;
```
kmh287, line 25: This seems better suited for Participant. Participant is a short-lived model that represents a user's single intention to find a match at a specific time. User should only contain data for a user that should live longer than any single attempt to match.

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -27,11 +46,15 @@ public UserDatastore(DatastoreService datastore) {
...
   public void addUser(User user) {
     // Set properties of entity based on user fields
-    Entity userEntity = new Entity(KIND_USER, user.getUsername());
-    userEntity.setProperty(PROPERTY_USERNAME, user.getUsername());
+    Entity entity = new Entity(KIND_USER, user.getUsername());
+    entity.setProperty(PROPERTY_USERNAME, user.getUsername());
+    entity.setProperty(PROPERTY_DURATION, user.getDuration());
+    entity.setProperty(PROPERTY_ROLE, user.getRole());
+    entity.setProperty(PROPERTY_PRODUCT_AREA, user.getProductArea());
+    entity.setProperty(PROPERTY_MATCH_PREFERENCE, user.getMatchPreference().getValue());
```
kmh287, line 54: WDYT about extracting a method to convert from a User to an entity, parallel to getUserFromEntity below?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -68,7 +71,19 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String role = formDetails.getString(REQUEST_ROLE);
     String productArea = formDetails.getString(REQUEST_PRODUCT_AREA);
     boolean savePreference = formDetails.getBoolean(REQUEST_SAVE_PREFERENCE);
-    String matchPreference = formDetails.getString(REQUEST_MATCH_PREFERENCE);
+    String matchPreferenceString = formDetails.getString(REQUEST_MATCH_PREFERENCE);
+    MatchPreference matchPreference = MatchPreference.ANY;
```
kmh287, line 75: Instead of having the default assigned here, it can be assigned in a default: branch of the switch along with any error logging for unexpected match preference Strings from the frontned.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: This method just says whether their preferences are compatible (only different + similar is not compatible). Then getCombinedMatchPreference finds overall what type of match both would want (similar + any = similar, different + any = different, similar + similar = similar, any + any = any, different + different = different). getSameNumFields says if half or more match and combinedMatchPreference = similar, then match.

jzwang8: Not sure if there's a better way to do this, so I'm definitely taking suggestions!

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Not all participants are ever added to the user datastore. They're only added to the datastore if the person opts to save preferences. And these fields are needed as part of a Participant if not all Participants have a corresponding user, unless we add every participant/user to user datastore and delete, but that would also require us to add a savePreference boolean/enum field to a Participant.

backend/src/main/java/com/google/sps/data/User.java
jzwang8: I was thinking that some people may prefer 30 min meetings over meetings of any other length, for example. That's the case for me at least. I'll ask about this in our pod meeting.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,77 @@ public Match findMatch(Participant newParticipant) {
...
-      if (compatibleTime) {
-        // Create and return match
-        return new Match(
-            newParticipant.getUsername(),
-            currParticipant.getUsername(),
-            duration,
-            currentTimeMillis);
+      if (combinedMatchPreference != MatchPreference.ANY) {
+        int numSameFields = getNumSameFields(newParticipant, currParticipant);
+        System.out.println(numSameFields);
```
tcwanggoogle, line 82: For logging purpose, please print more text on what this number means.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,77 @@ public Match findMatch(Participant newParticipant) {
...
+        if (combinedMatchPreference == MatchPreference.SIMILAR && numSameFields < MIN_SAME_FIELDS) {
+          continue;
+        } else if (combinedMatchPreference == MatchPreference.DIFFERENT
```
tcwanggoogle, line 85: No need for `else`

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,77 @@ public Match findMatch(Participant newParticipant) {
...
+            && numSameFields >= MIN_SAME_FIELDS) {
+          continue;
+        }
```
tcwanggoogle, line 88: To be consistent with `isCompatibleTime ` and `isCompatibleMatchPreference`, can we encapsulate this part into `isMatchPreferenceSatisfied(newMatchPreference, currMatchPreference)`?

backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Represent whether or not a Participant is matched */
+public enum MatchPreference {
+  /** Prefer to be matched with different Googler */
+  DIFFERENT(-1),
```
tcwanggoogle, line 20: In other languages, I think I usually see enum starts with 0, or all with positive numbers, so it's a little bit weird to see negative value here. Is there any reason to assign negative value here?

backend/src/main/java/com/google/sps/data/MatchStatus.java
```diff
@@ -18,7 +32,6 @@ public int getValue() {
   }
 
   public static MatchStatus forIntValue(int value) {
-    System.out.println(value);
```
tcwanggoogle, line 21: nit: In general, it's no harm to keep logs in your program for debugging purpose, so there's usually no need to remove a logging statement. (But I understand we don't use logging module but print to stdout, so removing might make sense here)

backend/src/main/java/com/google/sps/data/Participant.java
tcwanggoogle: +1 to Kevin's comment. I think the reason we use two different classes User and Participant is that `User` is used to describe a person, and `Participant` is used to indicate this person wants to participate a match. So we should only need to save the field to either one. As for opt not to save the preference, does our code support not adding `User` to datastore? If so, we can just create an object to save the input properties, but never add it to datastore.

backend/src/main/java/com/google/sps/FindMatchQuery.java
dhruvi618: +1

backend/src/main/java/com/google/sps/FindMatchQuery.java
dhruvi618: Agree with Kevin's comment. What do you think about combining both of the methods into one?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -92,17 +89,17 @@ public Match findMatch(Participant newParticipant) {
...
+          newParticipant.getUsername(), currParticipant.getUsername(), duration, clock.millis());
     }
     // No inital match found
     return null;
   }
 
   /** Return true if compatible endTimeAvailable considering duration and padding */
   private boolean isCompatibleTime(
-      long currentTimeMillis, int duration, long newEndTimeAvailable, long currEndTimeAvailable) {
+      int duration, long newEndTimeAvailable, long currEndTimeAvailable) {
```
kmh287, line 100: do `new` and `curr` make sense in the context of this method? Maybe something like participant1EndTime, participant2EndTime?

backend/src/main/java/com/google/sps/datastore/UserDatastore.java
```diff
@@ -53,8 +53,14 @@ public void addUser(User user) {
     entity.setProperty(PROPERTY_PRODUCT_AREA, user.getProductArea());
     entity.setProperty(PROPERTY_MATCH_PREFERENCE, user.getMatchPreference().getValue());
 
+    return entity;
+  }
+
+  /** Put user in datastore. Overwrite user entity if user with same username already exists. */
+  public void addUser(User user) {
+    ;
```
kmh287, line 61:
```suggestion
    
```

backend/src/main/java/com/google/sps/FindMatchQuery.java
kmh287: Got it. I think the method makes sense as-is but it could use javadoc that includes that explanation. Alternatively Dhruvi's suggestion to merge the methods could also be helpful; the method could return `null` if the two match preferences are incompatible.

backend/src/main/java/com/google/sps/data/MatchPreference.java
kmh287: +1 (literally!) please start with 0

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: changed new to first, curr to second

backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -0,0 +1,44 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Represent whether or not a Participant is matched */
```
TcgVanguardTroll, line 17: Is this  new class is meant to replace match status ?

TcgVanguardTroll: Or alongside it , this comment suggests the former

jzwang8: MatchStatus (whether or not a participant is matched yet) and MatchPreference (whether they want to be matched with someone similar, different, or anyone) are different.

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -68,7 +71,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        break;
+      case "any":
+        matchPreference = MatchPreference.ANY;
+        break;
+      case "similar":
+        matchPreference = MatchPreference.SIMILAR;
+        break;
+      default:
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid match preference.");
+        return;
```
TcgVanguardTroll, line 88: I like the use of switch cases !

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -112,6 +126,9 @@ public void compatibleTimeAndDuration() {
             TIME_1400ET,
             TIME_1600ET,
             DURATION_15_MINUTES,
+            ROLE_SOFTWARE_ENGINEER,
+            PRODUCT_AREA_ADS,
+            MATCH_PREFERENCE_SIMILAR,
             MATCHID_DEFAULT,
             MATCHSTATUS_UNMATCHED,
             TIMESTAMP_DEFAULT);
```
TcgVanguardTroll, line 134: I noticed that a lot of logic is repeated in created participates, do you think it would be good to extract some of it to a helper function ?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
```diff
@@ -368,6 +433,318 @@ public void barelyCompatible() {
...
+
+    assertThat(match.getFirstParticipantUsername()).isEqualTo(PERSON_B);
+    assertThat(match.getSecondParticipantUsername()).isEqualTo(PERSON_A);
+    assertThat(match.getDuration()).isEqualTo(DURATION_45_MINUTES);
+  }
+
+  @Test
+  public void oneFieldMatches() {
+    // Two participants with one matching field prefer to be matched with someone similar
+    // TODO: change num matching fields threshold when interests are added
```
TcgVanguardTroll, line 582: Nit : Is there a bug for this yet ?

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
jzwang8: What kind of helper function? I think it helps to see all of the fields in each participant so that debugging each case is easier.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,107 @@ public Match findMatch(Participant newParticipant) {
...
+    return (clock.millis() + TimeUnit.MINUTES.toMillis(duration + PADDING_MINUTES))
+        < earliestEndTimeAvailable;
+  }
+
+  /**
+   * @return combined MatchPreference similar = both are similar OR one is similar and one is any;
+   *     any = both are any; different = both are different = one is different and one is any. or
+   *     null if match preferences are not compatible
+   */
+  private MatchPreference getCompatibleMatchPreference(
```
dhruvi618, line 108: This looks good

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,107 @@ public Match findMatch(Participant newParticipant) {
...
+  /**
+   * @return combined MatchPreference similar: both are similar, one is similar and one is any any:
+   *     both are any different: both are different, one is different and one is any
+   */
+  private MatchPreference getCombinedMatchPreference(
+      MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
+    return (firstMatchPreference == MatchPreference.ANY)
+        ? secondMatchPreference
+        : firstMatchPreference;
+  }
```
dhruvi618, line 130: Since you refactored the methods to combine them, would this logic already be accounted for in `getCompatibleMatchPreference()`?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,107 @@ public Match findMatch(Participant newParticipant) {
...
+      // Count number of same fields between the two participants
+      boolean sameRole = newParticipant.getRole().equals(currParticipant.getRole());
+      boolean sameProductArea =
+          newParticipant.getProductArea().equals(currParticipant.getProductArea());
+      int numSameFields = Booleans.countTrue(sameRole, sameProductArea);
+      System.out.println("number of same fields: " + numSameFields);
+
+      // If match preference = similar, numSameFields must be at least MIN_SAME_FIELDS to satisfy
+      // preference
+      // If match preference = differe, numSameFields must be less than MIN_SAME_FIELDS to satisfy
```
dhruvi618, line 150: tiny nit: different

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: oops yeah, forgot to delete this! thanks!

backend/src/main/java/com/google/sps/data/MatchPreference.java
TcgVanguardTroll: Okay !

backend/src/test/java/com/google/sps/FindMatchQueryTest.java
TcgVanguardTroll: I do agree that as it stands right now it is very  blatant as to what the fields are,but, I believe a create Participant function just for this test suite would be fine ?.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
+
+      // If match preference = similar, numSameFields must be at least MIN_SAME_FIELDS to satisfy
+      // preference
+      // If match preference = different, numSameFields must be less than MIN_SAME_FIELDS to satisfy
+      // preference
+      if ((combinedMatchPreference == MatchPreference.SIMILAR && numSameFields < MIN_SAME_FIELDS)
+          || (combinedMatchPreference == MatchPreference.DIFFERENT
+              && numSameFields >= MIN_SAME_FIELDS)) {
+        return false;
+      }
```
tcwanggoogle, line 145: Same comment about logic here. I think the two parts of "||" can be broken into two ifs, with the comments are already written.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
-
-      if (compatibleTime) {
-        // Create and return match
-        return new Match(
-            newParticipant.getUsername(),
-            currParticipant.getUsername(),
-            duration,
-            currentTimeMillis);
+      long secondEndTimeAvailable = currParticipant.getEndTimeAvailable();
+      if (!isCompatibleTime(duration, firstEndTimeAvailable, secondEndTimeAvailable)) {
```
kmh287, line 66: Can you add a TODO + issue to push this into the datastore query? We should be able to query for all users with a matching duration %% whose endTimeAvailable is at least duration + PADDING minutes after now

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
+        System.out.println("not compatible time");
+        continue;
+      }
+
+      // Check match preference compatibility and get combined preference if compatible
+      MatchPreference secondMatchPreference = currParticipant.getMatchPreference();
+      MatchPreference combinedMatchPreference =
+          getCombinedMatchPreference(firstMatchPreference, secondMatchPreference);
+      if (combinedMatchPreference == null) {
+        System.out.println("not compatible match pref");
```
kmh287, line 76: replace print with comment?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
+      MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
+    if ((firstMatchPreference == secondMatchPreference)
+        || (firstMatchPreference == MatchPreference.ANY)
+        || (secondMatchPreference == MatchPreference.ANY)) {
+      return (firstMatchPreference == MatchPreference.ANY)
+          ? secondMatchPreference
+          : firstMatchPreference;
+    } else {
+      return null;
+    }
```
kmh287, line 118: This is a bit difficult to parse. WDYT about flipping the condition

```suggestion
  private MatchPreference getCombinedMatchPreference(
      MatchPreference firstMatchPreference, MatchPreference secondMatchPreference) {
    if ((firstMatchPreference != secondMatchPreference)
        && (firstMatchPreference != MatchPreference.ANY)
        && (secondMatchPreference != MatchPreference.ANY)) {
       	// One user is SIMILAR, other is DIFFERENT. Incompatible.
       	return null;
    }
    return (firstMatchPreference == MatchPreference.ANY)
          ? secondMatchPreference
          : firstMatchPreference;

```

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
+
+  /**
+   * @return true if the participants are a match based on their combinedMatchPreference and the
+   *     number of fields that are the same, false if not a match
+   */
+  private boolean isCombinedMatchPreferenceSatisfied(
+      MatchPreference combinedMatchPreference,
+      Participant newParticipant,
+      Participant currParticipant) {
+    if (combinedMatchPreference != MatchPreference.ANY) {
```
kmh287, line 129: flip this condition for an early return to unindent the following code by 2

```
if (combinedMatchPreference == MatchPreference.ANY) {
  // No need to check similarity
  return true;
}
...
```

backend/src/main/java/com/google/sps/data/User.java
kmh287: After the discussion sg to keep it. I would amend the comment to mention that this is a preference and not just related to a single matching attempt.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -50,30 +54,96 @@ public Match findMatch(Participant newParticipant) {
...
         participantDatastore.getParticipantsWithDuration(duration);
 
-    // Set reference date time using clock
-    long currentTimeMillis = clock.millis();
-
-    long newEndTimeAvailable = newParticipant.getEndTimeAvailable();
+    long firstEndTimeAvailable = newParticipant.getEndTimeAvailable();
+    MatchPreference firstMatchPreference = newParticipant.getMatchPreference();
+    String firstRole = newParticipant.getRole();
+    String firstProductArea = newParticipant.getProductArea();
```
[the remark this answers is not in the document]
jzwang8: You're right, these Strings aren't used but I do include the role/PA part in the matching algo in isCombinedMatchPreferenceSatisfied()

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
jzwang8: @kmh287 @tcwanggoogle You each suggested a different way to make this logic clearer. I ended up choosing Tiancong's here just because I think it makes the logic more straightforward. WDYT?

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: I changed it in this PR, but thanks for pointing this out. This was one of the reasons why we chose to only allow same durations to be matched.

backend/src/main/java/com/google/sps/FindMatchQuery.java
jzwang8: See the comment I tagged you in above in Tiancong's comment.

kmh287: Thanks I like Tiancong's suggestion better.

backend/src/main/java/com/google/sps/FindMatchQuery.java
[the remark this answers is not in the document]
tcwanggoogle: Yep, I understand. My point was the variable `firstRole` and `firstProductArea` are not used in this file, so we should not extract them.

PR #68, opened 2020-07-22
 Refactoring Email Notifier and Integrating it with Servlet 

TcgVanguardTroll:
This PR seeks to integrate our mailing system into our prototype.  This pull request required that the email notifier be refactored specifically changing our notifying methods to take strings that would represent recipients rather than storing them in a private data structure. This PR also creates the Gmail Factory Class that's instantaneously creates Gmail services that instantiate the Email notifier.

- `.gitignore`: Removes The folder containing binary folders 

- `GmailFactory.java `: This Class follows the Factory Builder method and is currently used within SearchMatchServlet in order to create the global instance of an email notifier .

- `EmailNotifier.java`: This File has undergone a refactoring, primarily removing a lot of the functionality that was delegated to other files, the notify user function was divided up into three scenarios based on search match servlet as well as delegating the authorization to another class. Currently, the class no longer has any private fields such as the recipients name as we felt they were better off as a method parameter 

- `EmailNotifierTest.java`: Due to the re-factoring of the Email Notifier class the tests required some changes as well as a few additional test cases . Currently the only additional tests ensure that in the cases of a noMatch, expired query , and match they have the correct subject .

- `SearchMatchServlet.java `: This File consists of the integration of the EmailNotifier. For each send .... response method a additional function call is used with a corresponding emailNotify ... send ... message. This change has led to the send response methods to be  given additional arguments primarily a Participant argument .


backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
    * @return the MimeMessage to be used to send email
    * @throws MessagingException if there was a problem accessing the Store
    */
-  private MimeMessage createEmail(String subject, String bodyText) throws MessagingException {
+  private MimeMessage createEmailWithSingleRecipient(
```
jzwang8, line 80: createEmailWithRecipient, or just createEmail is fine here (see my comment below about multi)

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+    email.addRecipients(
+        RecipientType.TO,
+        new InternetAddress[] {
+          new InternetAddress(firstRecipientUsername + "@google.com"),
+          new InternetAddress(secondRecipientUsername + "@google.com")
+        });
+    email.setSubject("Ad-lib: Match Found");
+    email.setText(bodyText);
+    return email;
+  }
```
jzwang8, line 121: No need for email to multiple recipients because each user in the match will get notifications at different times! (every 5 seconds)

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+
   /** Function that access its api and using it sends an email */
-  //   TODO(#35): Create a dummy email for ad lib itself to send emails.
   //   TODO(#36): Replace body to send real link to user instead of generic.
-  public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  public void sendMatchEmail(String firstMatchRecipientName, String secondMatchRecipientName)
```
jzwang8, line 125: Sending a match email to only one person, how about sendMatchEmail(String recipientUsername, String matchUsername). This would be called twice for each participant, and once in each doGet call.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+      throws MessagingException, IOException, GeneralSecurityException {
     MimeMessage email =
-        createEmail(
-            "Ad-Lib Meeting Found",
-            " Hey "
-                + recipientName
+        createEmailWithMultiRecipients(
+            firstMatchRecipientName,
+            secondMatchRecipientName,
+            "Congratulations"
```
jzwang8, line 131: Since don't have the Meet link yet, and may not since that requires getting that info from the Calendar part, how about just saying: 
```Hey recipientUsername!

We found you a match with matchUsername. Check your calendar for your meeting event, and feel free to join the Meet call now!

Best,
The Ad-lib team```

File: backend/src/main/java/com/google/sps/notifs/EmailNotifier.java [hunk 5]
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+    MimeMessage email = new MimeMessage(session);
+
+    email.setFrom(new InternetAddress(APPLICATION_EMAIL));
+    email.addRecipients(
+        RecipientType.TO,
+        new InternetAddress[] {
+          new InternetAddress(firstRecipientUsername + "@google.com"),
+          new InternetAddress(secondRecipientUsername + "@google.com")
+        });
+    email.setSubject("Ad-lib: Match Found");
```
jzwang8, line 118: [Ad-lib] We found you a match!

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
-  public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  public void sendMatchEmail(String firstMatchRecipientName, String secondMatchRecipientName)
+      throws MessagingException, IOException, GeneralSecurityException {
     MimeMessage email =
-        createEmail(
-            "Ad-Lib Meeting Found",
-            " Hey "
-                + recipientName
+        createEmailWithMultiRecipients(
```
jzwang8, line 128: Also make sure to set the subject when this multi method isn't used anymore.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+  public void sendExpiredEmail(String expiredRecipientName) throws MessagingException, IOException {
+    MimeMessage email =
+        createEmailWithSingleRecipient(
+            expiredRecipientName,
+            "Ad-lib: Meeting Query Expired",
+            " We apologize "
+                + expiredRecipientName
+                + "\n"
+                + " You're search for a match took longer than anticipated ! \n"
+                + " Please try again at a later time ! : \n");
```
jzwang8, line 149: subject: [Ad-lib] Sorry, we couldn't find you a match!
body: 
```Hi recipientUsername,

We couldn't find a match for you this time, but we encourage you to please try again later! 

Best,
The Ad-lib team```

File: backend/src/main/java/com/google/sps/notifs/EmailNotifier.java [hunk 8]
```diff
@@ -128,33 +77,93 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+        createEmailWithSingleRecipient(
+            noMatchParticipantName,
+            "Ad-lib: No Match Found",
+            " Sorry "
+                + noMatchParticipantName
+                + " We could'nt find you a Match :( \n"
+                + " Please try again later.");
+    Message messageWithEmail = createMessageWithEmail(noMatchEmail);
+    service.users().messages().send("me", messageWithEmail).execute();
+  }
```
jzwang8, line 168: The no match JSON response just tells the frontend that no match has been found YET so an email doesn't need to be sent here. Your correct that the expired JSON response needs to send an email though since this means we're taking that expired user out of the queue.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -43,9 +47,19 @@
   private static final String JSON_FIRST_PARTICIPANT_USERNAME = "firstParticipantUsername";
   private static final String JSON_SECOND_PARTICIPANT_USERNAME = "secondParticipantUsername";
   private static final String JSON_DURATION = "duration";
+  private final GmailFactory GM_Factory = new GmailFactory();
+  private final EmailNotifier EMAIL_NOTIFIER = new EmailNotifier(GM_Factory.build());
```
jzwang8, line 51: Add javadocs here?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,20 +125,30 @@ private boolean isExpired(Participant participant) {
...
+  private void sendNoMatchResponse(HttpServletResponse response, Participant participant)
+      throws IOException {
     JSONObject noMatchYet = new JSONObject();
     noMatchYet.put(JSON_MATCH_STATUS, "false");
-
+    try {
+      EMAIL_NOTIFIER.sendNoMatchEmail(participant.getUsername());
+    } catch (MessagingException e) {
+      e.printStackTrace();
+    }
```
jzwang8, line 151: No email necessary here!

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,7 +161,12 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
     matchExists.put(JSON_FIRST_PARTICIPANT_USERNAME, match.getFirstParticipantUsername());
     matchExists.put(JSON_SECOND_PARTICIPANT_USERNAME, match.getSecondParticipantUsername());
     matchExists.put(JSON_DURATION, match.getDuration());
-
+    try {
+      EMAIL_NOTIFIER.sendMatchEmail(
+          match.getFirstParticipantUsername(), match.getSecondParticipantUsername());
+    } catch (MessagingException | GeneralSecurityException e) {
+      e.printStackTrace();
+    }
```
jzwang8, line 169: Send email to only the current participant (could be firstParticipant or secondParticipant)! But also yes, pass in the other participant's username (have to figure out if the participant is first or second)

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -40,7 +40,7 @@
 
   @Before
   public void setUp() throws IOException {
-    testEmail = "jdoe@gmail.com";
+    testEmail = "John@google.com";
```
jzwang8, line 43: Maybe set testEmail to your own email?

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,63 @@
...
+public class GmailFactory {
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final String APPLICATION_NAME = "Ad-lib";
+  private static final String CREDENTIALS_FILE_PATH = "backend/credentials.json";
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  private final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+
+  public GmailFactory() throws GeneralSecurityException, IOException {}
+  ;
```
jzwang8, line 32: delete the ";"?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
jzwang8, line 51: make this camelcase?

TcgVanguardTroll: Done

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Ack

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Although it's not  necessary, what are we doing in the case of no match ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ahh, Okay

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Commen above explained

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack, added !

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: i am unclear on as to what you mean, the recipient username would refer to who is going to receive this match email but the match username is going to be who they are matched with ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Should I add a TODO

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Deleted Function

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: I believe I addressed this !

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,69 @@
...
+    GoogleClientSecrets clientSecrets =
+        GoogleClientSecrets.load(JSON_FACTORY, new InputStreamReader(in));
+
+    // Build flow and trigger user authorization request.
+    GoogleAuthorizationCodeFlow flow =
+        new GoogleAuthorizationCodeFlow.Builder(HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES)
+            .setDataStoreFactory(new FileDataStoreFactory(new File(TOKENS_DIRECTORY_PATH)))
+            .setAccessType("offline")
+            .build();
+    LocalServerReceiver receiver = new LocalServerReceiver.Builder().setPort(8000).build();
```
tcwanggoogle, line 60: nit: Maybe this port can be a const too? I don't see the need to change it but just to keep consistent with other consts

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.security.GeneralSecurityException;
+import java.util.List;
+
+public class GmailFactory {
```
kmh287, line 24: This class looks great. This is exactly what I had in mind. 👍

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,69 @@
...
+    // Load client secrets.
+    InputStream in = new FileInputStream(CREDENTIALS_FILE_PATH);
+    GoogleClientSecrets clientSecrets =
+        GoogleClientSecrets.load(JSON_FACTORY, new InputStreamReader(in));
+
+    // Build flow and trigger user authorization request.
+    GoogleAuthorizationCodeFlow flow =
+        new GoogleAuthorizationCodeFlow.Builder(HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES)
+            .setDataStoreFactory(new FileDataStoreFactory(new File(TOKENS_DIRECTORY_PATH)))
+            .setAccessType("offline")
```
kmh287, line 58: please extract this string constant. Also what are the other possible access types?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -47,60 +29,17 @@
 /** Class representing the notification system capable of sending mail to the user. */
 public class EmailNotifier {
 
-  private static final String APPLICATION_NAME = "Ad-lib";
+  //   TODO(#35): Create a dummy email for ad lib itself to send emails.
```
kmh287, line 32: email account / email address?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,32 +67,47 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
    * @return the MimeMessage to be used to send email
    * @throws MessagingException if there was a problem accessing the Store
    */
-  private MimeMessage createEmail(String subject, String bodyText) throws MessagingException {
+  private MimeMessage createEmail(String recipientUsername, String subject, String bodyText)
```
kmh287, line 70: How about recipientEmail? `username` sounds like it could be connected to our `User` model.

kmh287, line 70: iny nit: `subjectText` and `bodyText` or `subject` and `body`

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,32 +67,47 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+      throws MessagingException {
 
     Properties props = new Properties();
     Session session = Session.getDefaultInstance(props, /* authenticator= */ null);
 
     MimeMessage email = new MimeMessage(session);
 
     email.setFrom(new InternetAddress(APPLICATION_EMAIL));
-    email.addRecipient(RecipientType.TO, new InternetAddress(recipientEmail));
+    email.addRecipient(RecipientType.TO, new InternetAddress(recipientUsername + "@google.com"));
```
kmh287, line 79: is there any way we can avoid baking the @google.com assumption into this class? For instance could we get the email address from the authenticated user in the servlet and then pass the full email address to this method?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,8 +48,20 @@
...
   private static final String JSON_DURATION = "duration";
 
+  // Create GmailFactory and use it to instantiate Email Notifier
+  private final GmailFactory gmFactory = new GmailFactory();
+  private final EmailNotifier emailNotifier = new EmailNotifier(gmFactory.build());
+
+  // Get DatastoreService and instiate Match and Participant Datastores
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final MatchDatastore matchDatastore = new MatchDatastore(datastore);
+  private final ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
```
kmh287, line 58: nice!

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  private static final List<String> SCOPES = ImmutableList.of(GmailScopes.MAIL_GOOGLE_COM);
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final String APPLICATION_NAME = "Ad-lib";
+  private static final String CREDENTIALS_FILE_PATH = "backend/credentials.json";
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  private final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+
+  public GmailFactory() throws GeneralSecurityException, IOException {}
```
kmh287, line 38: which of the fields needs these `throws` clauses on the constructor?

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Creates an authorized Credential object.
+   *
+   * @param HTTP_TRANSPORT The network HTTP Transport.
+   * @return An authorized Credential object.
+   * @throws IOException If the credentials.json file cannot be found.
+   */
+  private static Credential getCredentials(final NetHttpTransport HTTP_TRANSPORT)
```
kmh287, line 47: nit: no need to mark the parameter as file. In older versions of java this was a useful hint to the compiler that the variable value shouldn't change. In more modern versions the compiler can determine that this is effectively final since it doesn't change int he body of the method.

backend/src/main/java/com/google/sps/GmailFactory.java
tcwanggoogle, line 38: Question: Does Java implicit constructor so you don't have to declare an empty one like this? Also, will it throw any exceptions?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,32 +67,47 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
+      throws MessagingException, IOException {
+    MimeMessage email =
+        createEmail(
+            firstMatchRecipientName,
+            "Ad-lib: We found you a match !",
+            " We found you a match with matchUsername. "
+                + " Check your calendar for your meeting event,"
+                + " and feel free to join the Meet call now!\n");
+    Message messageWithEmail = createMessageWithEmail(email);
+    service.users().messages().send("me", messageWithEmail).execute();
```
tcwanggoogle, line 97: What does this "me" means? Is it the sender? (I just want to make sure this PR can send email to the recipient or we expect a later one)

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,20 +127,25 @@ private boolean isExpired(Participant participant) {
...
+      e.printStackTrace();
+    }
     // Send the JSON back as the response
     response.setContentType("application/json");
     response.getWriter().println(expired.toString());
   }
 
   /** Send JSON response for no match yet */
-  private void sendNoMatchResponse(HttpServletResponse response) throws IOException {
+  private void sendNoMatchResponse(HttpServletResponse response, Participant participant)
```
tcwanggoogle, line 145: Do we need a sendNoMatchEmail() function for this one?

backend/src/main/java/com/google/sps/GmailFactory.java
dhruvi618: +1

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,32 +67,47 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
   //   TODO(#36): Replace body to send real link to user instead of generic.
-  public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  public void sendMatchEmail(String firstMatchRecipientName)
+      throws MessagingException, IOException {
+    MimeMessage email =
+        createEmail(
+            firstMatchRecipientName,
+            "Ad-lib: We found you a match !",
+            " We found you a match with matchUsername. "
```
dhruvi618, line 93: Is there a way we can get the email address/username from servlet Users API or even the Match object here to replace the "matchUsername" with an actual username instead?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -128,32 +67,47 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
...
     MimeMessage email =
         createEmail(
-            "Ad-Lib Meeting Found",
-            " Hey "
-                + recipientName
-                + " Please Join your Ad-Lib meeting via the link below : \n"
-                + " http://meet.google.com/new");
+            expiredRecipientName,
+            "Ad-lib: Sorry, we couldn't find you a match!",
+            " We couldn't find a match  for you this time, but we encourage you to please try "
```
dhruvi618, line 107: tiny nit: extra space between match and for

TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: I have changed it to use the string name parameter that has been passed into the function.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: No we do not !

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: "Me" refers to the authorized sender, it is a shortcut for me trying into "grantjustice". Currently, this PR is capable of sending emails however we do not have the ability to have anyone besides me send emails

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: That makes sense

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: The HTTP_TRANSPORT can cause these errors

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: email account

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Extracted, there are three access types, Online, offline, and null , "online" to request online access or "offline" to request offline access or null for the default behavior of "online"

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Thanks

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Removed the @google.com prefix !

backend/src/main/java/com/google/sps/GmailFactory.java
kmh287: why do we request offline access?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
tcwanggoogle: Thanks. Just want to make sure I understand it. Using your email is perfectly fine atm.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
tcwanggoogle: If we don't need to send email, why do we need to pass Participant? i.e we don't need this change.

TcgVanguardTroll: Fixed

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Java does implicit constructor but due to the method calls within the class (get credentials) the constructor is used just for the sake of throwing errors which will occur via the HTTTP_Transport

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/GmailFactory.java
kmh287: I would discourage throwing checked exceptions in the constructor for exactly this reason. The throws clause is contagious and now the servlets also have to define a default constructor that throws these exceptions. WDYT about creating the transport where it's needed instead of creating ti at construction time?

TcgVanguardTroll: Okay, that would make sense

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,10 +128,16 @@ private boolean isExpired(Participant participant) {
...
   /** Send JSON response for expired participant that has been removed from datastore */
-  private void sendExpiredResponse(HttpServletResponse response) throws IOException {
+  private void sendExpiredResponse(HttpServletResponse response, Participant participant)
+      throws IOException {
     JSONObject expired = new JSONObject();
     expired.put(JSON_MATCH_STATUS, "expired");
-
+    try {
+      emailNotifier.sendExpiredEmail(
+          participant.getUsername(), userService.getCurrentUser().getEmail());
```
jzwang8, line 137: Passing participant in as a parameter passes unnecessary info. You could pass just the username (not participant.getUsername(), but the String username from line 75) and either append the "@google.com" when passing email to sendExpiredEmail here OR just pass the username and append the "@google.com" in the sendExpiredEmail() method in EmailNotifier

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,7 +159,12 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
     matchExists.put(JSON_FIRST_PARTICIPANT_USERNAME, match.getFirstParticipantUsername());
     matchExists.put(JSON_SECOND_PARTICIPANT_USERNAME, match.getSecondParticipantUsername());
     matchExists.put(JSON_DURATION, match.getDuration());
-
+    try {
+      emailNotifier.sendMatchEmail(
+          match.getFirstParticipantUsername(), userService.getCurrentUser().getEmail());
```
jzwang8, line 164: same comment here as above in sendExpiredResponse()

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,23 +78,41 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
         createEmail(
-            "Ad-Lib Meeting Found",
-            " Hey "
-                + recipientName
-                + " Please Join your Ad-Lib meeting via the link below : \n"
-                + " http://meet.google.com/new");
+            expiredRecipientEmail,
+            "Ad-lib: Sorry, we couldn't find you a match!",
+            String.format(
+                " Sorry %s We couldn't find a match for you this time, but we encourage you to "
```
jzwang8, line 111: What's the %s?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,23 +78,41 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
-  //   TODO(#35): Create a dummy email for ad lib itself to send emails.
   //   TODO(#36): Replace body to send real link to user instead of generic.
-  public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  public void sendMatchEmail(String firstMatchRecipientName, String firstMatchRecipientEmail)
+      throws MessagingException, IOException {
+    MimeMessage email =
+        createEmail(
+            firstMatchRecipientEmail,
+            "Ad-lib: We found you a match !",
```
jzwang8, line 93: Remove space after "match"

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,23 +78,41 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
+            String.format(
+                " We found you a match with %s "
+                    + " Check your calendar for your meeting event,"
+                    + " and feel free to join the Meet call now!\n",
+                firstMatchRecipientName));
+    Message messageWithEmail = createMessageWithEmail(email);
+    service.users().messages().send("me", messageWithEmail).execute();
```
jzwang8, line 100: WDYT about extracting "me" as a string constant with a javadoc so we know what this means and what to replace this with if we ever did get a different authorized sender?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: It is just a regular expression to express that there will be a string, the string is the second argument in the String.format function which is the recipient username

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Kevin had recommended against adding @google.com to the strings . Also in the case in which i get the username form line 75 should username be set to a class field ?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,24 +48,32 @@
...
     System.out.println("Request received");
 
-    // Get participant username
-    UserService userService = UserServiceFactory.getUserService();
     String email = userService.getCurrentUser().getEmail();
     if (email == null) {
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
       return;
     }
     String username = email.split("@")[0];
```
dhruvi618, line 75: WDYT about potentially extracting this out into its own method, so that the doGet becomes a bit clearer to read?

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: That would be useful !

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: I am confused as to what you mean in this case.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: dhruvi mentioned extracting the logic form line 70-75 so I will try to go about doing that !

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,24 +48,32 @@
...
+
+  // Get DatastoreService and instiate Match and Participant Datastores
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final MatchDatastore matchDatastore = new MatchDatastore(datastore);
+  private final ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
+
+  // Get participant username
+  UserService userService = UserServiceFactory.getUserService();
+
+  public SearchMatchServlet() throws GeneralSecurityException, IOException {}
```
kmh287, line 63: no longer needed :)

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final int PORT_NUM = 8000;
+
+  /**
+   * Creates an authorized Credential object.
+   *
+   * @param HTTP_TRANSPORT The network HTTP Transport.
+   * @return An authorized Credential object.
+   * @throws IOException If the credentials.json file cannot be found.
+   */
+  private static Credential getCredentials(NetHttpTransport HTTP_TRANSPORT) throws IOException {
```
kmh287, line 46: function parameters should be camelCased

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
dhruvi618: What if we added a getUsernameFromEmail method that would take in the email which we retrieve from the userService here and just returns `email.split("@")[0]`? Then, I think we would only need the email param since the username can be retrieved from the email

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,68 @@
...
+        new GoogleAuthorizationCodeFlow.Builder(HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES)
+            .setDataStoreFactory(new FileDataStoreFactory(new File(TOKENS_DIRECTORY_PATH)))
+            .setAccessType(ACCCESS_TYPE)
+            .build();
+    LocalServerReceiver receiver = new LocalServerReceiver.Builder().setPort(PORT_NUM).build();
+    return new AuthorizationCodeInstalledApp(flow, receiver).authorize("user");
+  }
+
+  public Gmail build() throws IOException, GeneralSecurityException {
+    final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
kmh287, line 63: Even without the final keyword Java can tell this is effectively final. Also please update the casing to camelCase.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,23 +78,41 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
     email.setText(bodyText);
     return email;
   }
 
   /** Function that access its api and using it sends an email */
-  //   TODO(#35): Create a dummy email for ad lib itself to send emails.
   //   TODO(#36): Replace body to send real link to user instead of generic.
-  public void notifyUser() throws MessagingException, IOException, GeneralSecurityException {
-    NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  public void sendMatchEmail(String firstMatchRecipientName, String firstMatchRecipientEmail)
```
kmh287, line 88: recipientName, recipientEmail now that "first" doesn't differentiate two recipients.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
kmh287: Absolutely right, though it's a string placeholder and not a regular expression.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
jzwang8: @dhruvi618 i know what you mean, I can do this part

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: i know that constructor sin java are implicit so empty constructors are handled by the JVM but since the class method that is being thrown throws error how are they handled if i remove this constructor or should instead of being exceptions to the build method should try catches be used ?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: i have extracted a get username that uses this logic If that is what you mean ?

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: @kmh287

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: @kmh287 ?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
kmh287: The email notifier should take in the full email address rather than construct it internally from a username + "@google.com" which assumes the user's email domain. `email` in the calling method has the email address, so how about passing that instead of the participant

TcgVanguardTroll: I have added a getter method similar to that of addPartcipantServlet and call that instead of taking a participant in the constructor

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
kmh287: I would recommend having the build method just throw the exception to the caller so the caller can handle them. Either way we should avoid throwing in the constructor.

TcgVanguardTroll: The current build method does throw in the caller ,however, i get a Unhandled exceptions: java.io.IOException, java.security.GeneralSecurityException on line 53 because it is unhandled

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,38 +48,41 @@
...
   private static final String JSON_DURATION = "duration";
 
+  // Create GmailFactory and use it to instantiate Email Notifier
+  private final GmailFactory gmFactory = new GmailFactory();
+  // Get DatastoreService and instiate Match and Participant Datastores
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final MatchDatastore matchDatastore = new MatchDatastore(datastore);
+  private final ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
+  // Get participant username
+  UserService userService = UserServiceFactory.getUserService();
```
kmh287, line 58: private final

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -123,12 +67,14 @@ private static Message createMessageWithEmail(MimeMessage emailContent)
   /**
    * Create a MimeMessage using the parameters provided.
    *
-   * @param subject subject of the email
+   * @param recipientEmail email of user to whom is to be notified
```
kmh287, line 70: email to which the notification should be sent ?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,38 +48,41 @@
...
 
-    // Get participant username
-    UserService userService = UserServiceFactory.getUserService();
-    String email = userService.getCurrentUser().getEmail();
-    if (email == null) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
-      return;
+    try {
+      emailNotifier = ensureEmailNotifier();
+    } catch (GeneralSecurityException e) {
```
kmh287, line 68: Is this used outside of sendExpiredResponse? If not can this block move into that method?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +153,24 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
     // Send the JSON back as the response
     response.setContentType("application/json");
     response.getWriter().println(matchExists.toString());
   }
+
+  /** Retrieve user email address via Users API and parse for username */
+  private String getUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
```
kmh287, line 170: when will email be null?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +153,24 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+  }
+
+  private EmailNotifier ensureEmailNotifier() throws GeneralSecurityException, IOException {
+    return this.emailNotifier != null ? this.emailNotifier : new EmailNotifier(gmFactory.build());
```
kmh287, line 174: If this method ends up instantiating a new EmailNotifier, let's also assign it to this.emailNotifier within this method

backend/src/test/java/com/google/sps/EmailNotifierTest.java
```diff
@@ -68,62 +67,69 @@ public MimeMessage convertToMimeMessage(Message msg) throws MessagingException {
   }
 
   @Test
-  public void testMessageHasCorrectApplicationName()
-      throws MessagingException, IOException, GeneralSecurityException {
+  public void testMessageHasCorrectApplicationName() throws MessagingException, IOException {
```
kmh287, line 70: here too

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: It is used for both send expired and sendMatch, so I'll move the logic to those two places !

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: The user doesn't exists

kmh287: Can that happen normally?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,10 +123,15 @@ private boolean isExpired(Participant participant) {
...
+  private void sendExpiredResponse(HttpServletResponse response, Participant participant)
+      throws IOException {
     JSONObject expired = new JSONObject();
     expired.put(JSON_MATCH_STATUS, "expired");
-
+    try {
+      emailNotifier.sendExpiredEmail(getUsername(), userService.getCurrentUser().getEmail());
+    } catch (MessagingException e) {
+      e.printStackTrace();
+    }
```
kmh287, line 134: Ok so now if there's a problem constructing an email notifier here, how should the servlet behave? Is it okay to continue on without sending an email? @jzwang8 @dhruvi618

backend/src/test/java/com/google/sps/EmailNotifierTest.java
TcgVanguardTroll: What does here too refer to exactly ?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: I'm not too familiar with the backend but I don't believe so as an error will be thrown if such field is null form what i've seen

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,24 +83,42 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
-                + " http://meet.google.com/new");
+            recipientEmail,
+            "Ad-lib: We found you a match!",
+            String.format(
+                " We found you a match with %s "
+                    + " Check your calendar for your meeting event,"
+                    + " and feel free to join the Meet call now!\n",
+                recipientName));
     Message messageWithEmail = createMessageWithEmail(email);
     service.users().messages().send("me", messageWithEmail).execute();
```
tcwanggoogle, line 105: Change this `me` to `AUTH_USERNAME` since we've already created a variable

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -44,38 +48,35 @@
...
-    if (email == null) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid email.");
-      return;
-    }
-    String username = email.split("@")[0];
-
-    // Get DatastoreService and instiate Match and Participant Datastores
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    MatchDatastore matchDatastore = new MatchDatastore(datastore);
-    ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
```
tcwanggoogle, line 63: We should not remove these lines.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,33 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
   }
+
+  /** Retrieve user email address via Users API and parse for username */
+  private String getUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
+
+  private EmailNotifier ensureEmailNotifier() throws GeneralSecurityException, IOException {
```
tcwanggoogle, line 177: Maybe rename the name to `createOrGetEmailNotifier()`? "ensure" is not clear to me

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,33 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+    if (this.emailNotifier == null) return null;
+    else {
```
tcwanggoogle, line 179: No else needed after return.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,33 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+
+  /** Retrieve user email address via Users API and parse for username */
+  private String getUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
+
+  private EmailNotifier ensureEmailNotifier() throws GeneralSecurityException, IOException {
+    if (this.emailNotifier == null) return null;
```
tcwanggoogle, line 178: Is this logic right? I think we want to create a new emailNotifier when it's null and return it when it's not?

TcgVanguardTroll: yeah it is sorry i was trying to copy the logic from a ternary operator standpoint

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
dhruvi618: I think Jordan added these datastore instantiations as fields in this servlet now, so I think these might be fine to remove

TcgVanguardTroll: I didn't remove them i  made them global

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: check line 54

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
dhruvi618: This will only be null if the Users API returns null (can't retrieve email). Since I do not think this can normally happen, I think we can just return the email the Users API returns and move forward from there. Otherwise, if we choose to keep the null return value from this method, what do you think about adding a null check to the places where this method is called elsewhere in the servlet?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
dhruvi618: If there's a problem constructing an email notifier here, I think it would okay to continue on without sending an email if the servlet could still return the JSON object to the frontend so we can notify the user that way

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: quoted from https://developers.google.com/api-client-library/java/google-api-java-client/oauth2 

"An access token typically has an expiration date of 1 hour, after which you will get an error if you try to use it. GoogleCredential takes care of automatically "refreshing" the token, which simply means getting a new access token. This is done by means of a long-lived refresh token, which is typically received along with the access token if you use the access_type=offline parameter during the authorization code flow "

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Adding null checking sounds like it could be fine !

backend/src/main/java/com/google/sps/GmailFactory.java
```diff
@@ -0,0 +1,70 @@
...
+            .setDataStoreFactory(new FileDataStoreFactory(new File(TOKENS_DIRECTORY_PATH)))
+            .setAccessType(ACCESS_TYPE)
+            .build();
+    LocalServerReceiver receiver = new LocalServerReceiver.Builder().setPort(PORT_NUM).build();
+    return new AuthorizationCodeInstalledApp(flow, receiver).authorize(AUTH_USER);
+  }
+
+  public Gmail build() throws GeneralSecurityException, IOException {
+    NetHttpTransport httpTransport;
+    httpTransport = GoogleNetHttpTransport.newTrustedTransport();
```
kmh287, line 65: Please join declaration and assignment 

`NetHttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();`

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,10 +117,20 @@ private boolean isExpired(Participant participant) {
...
+    try {
+      emailNotifier = createOrGetEmailNotifier();
+    } catch (GeneralSecurityException e) {
+      e.printStackTrace();
+    }
+    try {
+      emailNotifier.sendExpiredEmail(getUsername(), userService.getCurrentUser().getEmail());
+    } catch (MessagingException e) {
+      e.printStackTrace();
+    }
```
kmh287, line 133: Join these together with multicatch 
https://stackoverflow.com/questions/3495926/can-i-catch-multiple-java-exceptions-in-the-same-catch-clause

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,31 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+    try {
+      emailNotifier = createOrGetEmailNotifier();
+    } catch (GeneralSecurityException e) {
+      e.printStackTrace();
+    }
+    try {
+      emailNotifier.sendMatchEmail(getUsername(), userService.getCurrentUser().getEmail());
+    } catch (MessagingException e) {
+      e.printStackTrace();
+    }
```
kmh287, line 164: can this be factored out of sendExpiredResponse and sendMatchResponse into a common method?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,31 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
     // Send the JSON back as the response
     response.setContentType("application/json");
     response.getWriter().println(matchExists.toString());
   }
+
+  /** Retrieve user email address via Users API and parse for username */
```
kmh287, line 170: include a description of when this method returns null

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,31 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+  private String getUsername() {
```
kmh287, line 171: Once thing we've started to use in other parts of the code are nullability annotations. Adding the @Nullable annotation to this method lets callers know that this method can return null. Even better, it hints to static analysis tools like linters that the return value could be null so they'll warn you if you try to dereference getUsername() without a null check. Please add `@Nullable`

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,31 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+    UserService userService = UserServiceFactory.getUserService();
```
kmh287, line 172: could this method use the existing `userService` field instead?

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +152,31 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
+
+  private EmailNotifier createOrGetEmailNotifier() throws GeneralSecurityException, IOException {
+    if (this.emailNotifier != null) return this.emailNotifier;
```
kmh287, line 178: Please add brackets around the return.
https://google.github.io/styleguide/javaguide.html#s4.1-braces

backend/src/test/java/com/google/sps/EmailNotifierTest.java
kmh287: My bad, I think I misread this when I left the comment.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ahh, yup ! Perks of making it global

backend/src/main/java/com/google/sps/GmailFactory.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: I believe it can be

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Done

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +144,45 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
+
+  private EmailNotifier createOrGetEmailNotifier() throws GeneralSecurityException, IOException {
+    if (this.emailNotifier != null) {
+      return this.emailNotifier;
+    }
+    this.emailNotifier = new EmailNotifier(gmFactory.build());
+    return this.emailNotifier;
```
kmh287, line 168: tiny nit: no need for `this` here

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -142,9 +144,45 @@ private void sendMatchResponse(HttpServletResponse response, Match match) throws
...
+      }
+    } else {
+      try {
+        emailNotifier = createOrGetEmailNotifier();
+        emailNotifier.sendExpiredEmail(getUsername(), userService.getCurrentUser().getEmail());
+      } catch (GeneralSecurityException | MessagingException | IOException e) {
+        e.printStackTrace();
+      }
+    }
+  }
```
kmh287, line 187: Hmmm, when I suggested this I didn't notice that the method calls on emailNotifier are different. Since they are I'd go further and replace this method with two methods, sendMatchEmail and sendExpiredEmail. A method that takes a boolean parameter that is used to switch between two different branches a sign that the method should be split in two.

TcgVanguardTroll: Ack split the function into two

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -116,10 +118,11 @@ private boolean isExpired(Participant participant) {
   }
 
   /** Send JSON response for expired participant that has been removed from datastore */
-  private void sendExpiredResponse(HttpServletResponse response) throws IOException {
+  private void sendExpiredResponse(HttpServletResponse response, Participant participant)
```
tcwanggoogle, line 121: It looks like `participant` is not used in this function. Please remove it.

TcgVanguardTroll: Ack

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -47,60 +29,22 @@
 /** Class representing the notification system capable of sending mail to the user. */
 public class EmailNotifier {
 
-  private static final String APPLICATION_NAME = "Ad-lib";
+  //   TODO(#35): Create a dummy email account for ad lib itself to send emails.
   private static final String APPLICATION_EMAIL = "Adlib-Step@gmail.com";
```
jzwang8, line 33: Change this to your email? I feel like a random email address shouldn't be in our public repo.

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
```diff
@@ -137,24 +83,42 @@ private MimeMessage createEmail(String subject, String bodyText) throws Messagin
...
         createEmail(
-            "Ad-Lib Meeting Found",
-            " Hey "
-                + recipientName
-                + " Please Join your Ad-Lib meeting via the link below : \n"
-                + " http://meet.google.com/new");
+            expiredRecipientEmail,
+            "Ad-lib: Sorry, we couldn't find you a match!",
+            String.format(
+                " Sorry %s We couldn't find a match for you this time, but we encourage you to "
```
jzwang8, line 116: "Sorry %s, we couldn't..."

backend/src/main/java/com/google/sps/notifs/EmailNotifier.java
TcgVanguardTroll: Okay, I've changed adlib-Step@gmail.com to my ldap

PR #74, opened 2020-07-23
Add interests to matching alg

jzwang8:
Files changed:
- FindMatchQuery: Add interests to alg, two participants considered similar if half of the max number of filled interests (rounded up) are the same, test
- Participant/ParticipantDatastore/ParticipantDatastoreTest: Add interests field and test
- User/UserDatastore/UserDatastoreTest: Add interests field and test
- AddParticipantServlet: Get and parse JSON interests


backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -47,48 +44,44 @@ public FindMatchQuery(Clock clock, ParticipantDatastore participantDatastore) {
...
       }
 
       // Check if combined match preference is satisfied depending on number of same fields
       if (!isCombinedMatchPreferenceSatisfied(
-          combinedMatchPreference, newParticipant, currParticipant)) {
+          combinedMatchPreference, firstParticipant, secondParticipant)) {
         continue;
       }
 
+      System.out.println("match found");
```
tcwanggoogle, line 73: As a log, it's better printing out "match found xx and yy"

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -114,29 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
+        }
+      }
+    }
+    System.out.println("numSameFields: " + numSameFields);
 
-    // If match preference = SIMILAR, numSameFields must be at least MIN_SAME_FIELDS to satisfy
-    // preference
+    // Check if match based on preference and number of same fields
+    int maxNumFilledFields = Math.max(firstNumFilledFields, secondNumFilledFields);
+    int MIN_SAME_FIELDS = maxNumFilledFields / 2 + maxNumFilledFields % 2;
```
tcwanggoogle, line 143: nit: (maxNumFilledFields + 1) / 2 should give you the same result

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -114,29 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
-    // If match preference = DIFFERENT, numSameFields must be less than MIN_SAME_FIELDS to satisfy
-    // preference
     if (combinedMatchPreference == MatchPreference.DIFFERENT && numSameFields >= MIN_SAME_FIELDS) {
       return false;
     }
+    System.out.println("combined match pref satisfied");
     return true;
   }
+
+  /** Return all filled fields of participant in one string delimited by , */
```
tcwanggoogle, line 155: It's worth a mention here that this function assumes none of the Role, PA or interests should have the same words in other categories. Otherwise, this match might fail

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -114,29 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
+    int firstNumFilledFields = firstFields.countTokens();
+    int secondNumFilledFields = secondFields.countTokens();
+    int numSameFields = 0;
+    while (firstFields.hasMoreTokens()) {
+      String firstField = firstFields.nextToken();
+      secondFields = new StringTokenizer(secondFilledFieldsList, ",");
+      while (secondFields.hasMoreTokens()) {
+        if (firstField.equals(secondFields.nextToken())) {
+          numSameFields++;
+          System.out.println(firstField + " matches");
```
tcwanggoogle, line 135: We should `break` here, though keep going won't give you a different answer.

backend/src/main/java/com/google/sps/FindMatchQuery.java
tcwanggoogle, line 143: I guess all capitals are reserved for consts and it's no longer a const anymore.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -114,29 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
+    System.out.println("combined match pref satisfied");
     return true;
   }
+
+  /** Return all filled fields of participant in one string delimited by , */
+  private String getAllFilledFieldsList(Participant participant) {
+    StringBuilder sb = new StringBuilder();
+    String role = participant.getRole();
+    boolean isRoleFilled = !role.equals("");
+    if (isRoleFilled) {
```
tcwanggoogle, line 160: nit: `if (!role.equals(""))` as it doesn't increase the length of the line and is clearer for people to understand the condition. Same comment below.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -114,29 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
     }
 
-    // Count number of same fields between the two participants
-    boolean sameRole = newParticipant.getRole().equals(currParticipant.getRole());
-    boolean sameProductArea =
-        newParticipant.getProductArea().equals(currParticipant.getProductArea());
-    int numSameFields = Booleans.countTrue(sameRole, sameProductArea);
-    System.out.println("number of same fields: " + numSameFields);
+    // Get lists of combined role, product area, interests for first and second participant
+    String firstFilledFieldsList = getAllFilledFieldsList(firstParticipant);
```
dhruvi618, line 118: What do you think about renaming this to firstParticipantInputs or something similar? Also similar comment about the method name, would it be possible to rename it to something without "List" at the end since it's returning a String ( maybe something like getAllParticipantInputs)?

backend/src/main/java/com/google/sps/data/Participant.java
```diff
@@ -32,6 +32,8 @@
   private final String role;
   /** Product area at Google */
   private final String productArea;
+  /** List of interests delimited by commas */
+  private final String interests;
```
kmh287, line 36: The comma separated String is a good storage format but a List of Strings would be easier to work with in Java. Also converting between List<String> and comma-separated String in participant datastore prevents the rest of the code from depending on an implementation detail of the storage

jzwang8:
> Don't forget to update the list of files

All 9 changed files are mentioned in my PR description? Is that okay? Because Participant/ParticipantDatastore/ParticipantDatastoreTest just involved adding an interests field.

kmh287:
My bad, description LG!

backend/src/main/java/com/google/sps/data/Participant.java
jzwang8: Added a ListStringConversion class with static methods to easily convert between List for Participant/User and String for their datastore representations!

backend/src/main/java/com/google/sps/datastore/ListStringConversion.java
```diff
@@ -0,0 +1,41 @@
...
+public class ListStringConversion {
+
+  /** Convert list of strings to a string with each element delimited by a comma */
+  public static String listToString(List<String> list) {
+    StringBuilder sb = new StringBuilder();
+    for (String str : list) {
+      sb.append(str + ",");
+    }
+    return sb.toString();
+  }
```
kmh287, line 30: https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

backend/src/main/java/com/google/sps/datastore/ListStringConversion.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /** Convert a string with each element delimited by a comma to a list of strings */
+  public static List<String> stringToList(String str) {
+    StringTokenizer st = new StringTokenizer(str, ",");
+    List<String> list = new ArrayList<String>();
+    while (st.hasMoreTokens()) {
+      list.add(st.nextToken());
+    }
+    return list;
+  }
```
kmh287, line 40: Arrays.asList(String.split(",", str))

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -153,8 +154,10 @@ public Participant getParticipantFromUsername(String username) {
 
     // Convert entities to list of participants
     List<Participant> participants = new ArrayList<Participant>();
-    for (Entity entity : results.asIterable()) {
-      participants.add(getParticipantFromEntity(entity));
+    if (results != null) {
```
kmh287, line 157: when is results null? when there are no entries found?

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -110,33 +107,65 @@ private MatchPreference getCombinedMatchPreference(
...
+    int numSameInputs = 0;
+    for (String firstFilledInput : firstFilledInputs) {
+      for (String secondFilledInput : secondFilledInputs) {
+        if (firstFilledInput.equals(secondFilledInput)) {
+          numSameInputs++;
+          System.out.println(firstFilledInput + "matches");
+          break;
+        }
+      }
+    }
```
kmh287, line 137: how about 
```
int numSameInputs = Sets.intersection(
    new HashSet<>(firstFilledInputs),
    new HashSet<>(secondFilledInputs)).size()
```

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
jzwang8: Oops, I deployed the backend and was debugging and forgot to remove this. Removed again.

backend/src/main/java/com/google/sps/datastore/ListStringConversion.java
```diff
@@ -14,28 +14,18 @@
 
 package com.google.sps.datastore;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
-import java.util.StringTokenizer;
 
 public class ListStringConversion {
```
kmh287, line 20: Now that these methods are shorter, WDYT about making them private static methods in the datastore class? 

It's ok if the test has to duplicate the logic to parse the String stored in the datastoreservice. The test is allowed to make some assumptions about implementation details of the unit under test.

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -47,48 +44,48 @@ public FindMatchQuery(Clock clock, ParticipantDatastore participantDatastore) {
    *     availibility, or null if no match yet
    */
   @Nullable
-  public Match findMatch(Participant newParticipant) {
-    int duration = newParticipant.getDuration();
+  public Match findMatch(Participant firstParticipant) {
```
kmh287, line 47: Unless a caller knows the implementation details of this method, `firstParticipant` could be a confusing parameter name. I like the previous `newMatch` or even just `participant`. To make the code in the method easier to read (where there is a second participant to think about) you could add

Participant firstParticipant = participant

backend/src/main/java/com/google/sps/FindMatchQuery.java
```diff
@@ -110,33 +107,57 @@ private MatchPreference getCombinedMatchPreference(
...
-    int numSameFields = Booleans.countTrue(sameRole, sameProductArea);
-    System.out.println("number of same fields: " + numSameFields);
+    // Get lists of combined role, product area, interests for first and second participant
+    List<String> firstFilledInputs = getAllFilledInputs(firstParticipant);
+    List<String> secondFilledInputs = getAllFilledInputs(secondParticipant);
+    System.out.println("First filled inputs: " + firstFilledInputs.toString());
+    System.out.println("Second filled inputs: " + secondFilledInputs.toString());
+
+    // Count number of shared inputs by finding size of intersection
+    (firstFilledInputs).retainAll(secondFilledInputs);
```
kmh287, line 128: tiny nit: no need for parens around `firstFilledInputs`

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -168,4 +172,14 @@ public void removeParticipant(String username) {
...
+
+  /** Convert list of strings to a string with each element delimited by a comma */
+  public static String convertListToString(List<String> list) {
+    return String.join(",", list);
+  }
+
+  /** Convert a string with each element delimited by a comma to a list of strings */
+  public static List<String> convertStringToList(String str) {
+    return Arrays.asList(str.split(","));
+  }
```
kmh287, line 184: private

jzwang8: I use these methods in ParticipantDatastoreTest. Should I add these methods to the test to make these private?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -127,14 +132,13 @@ public Participant getParticipantFromUsername(String username) {
 
   /** Return list of all unmatched participants with duration and compatible endTimeAvailable */
   public List<Participant> getParticipantsCompatibleTimeAvailibility(
-      int duration, long endTimeAvailable, int paddingTime, Clock clock) {
-    Query query = new Query(KIND_PARTICIPANT);
-
+      int duration, long endTimeAvailable, int paddingTime, Clock clock)
+      throws DatastoreNeedIndexException {
```
TcgVanguardTroll, line 136: When will this function throw this exception ?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -47,51 +50,55 @@
...
 
     // Get endTimeAvailable and startTimeAvailable in milliseconds
     long endTimeAvailable = formDetails.getLong(REQUEST_END_TIME_AVAILABLE);
     long startTimeAvailable = Instant.now().toEpochMilli();
 
+    // Get all participant inputs
     int duration = formDetails.getInt(REQUEST_DURATION);
     if (duration <= 0) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid duration.");
+      response.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Invalid duration.");
```
dhruvi618, line 79: Can we changed all these errors back to BAD_REQUEST errors?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
kmh287: That sounds good. This functionality shouldn't be part of the public API.

PR #75, opened 2020-07-23
Add GET request query to search-match servlet and corresponding unit test

dhruvi618:
Files changed: 

- App.js: Add Axios GET request query to servlet and change state of matchStatus based on the response and add props to Form component to establish communication between App and Form
- App.test.js: Add GET request test with mocked response
- Form.js: Add onSubmit props along with validation to Form and fix lint errors
- Form.test.js: Fix trailing space lint error

Snapshot files changed:
- App.test.js.snap: Add match status paragraph component to snapshot file


frontend/src/App.js
```diff
@@ -26,6 +27,24 @@ const useStyles = makeStyles((theme) => ({
...
+    axios.get('/api/v1/search-match')
+        .then((response) => {
+          console.log(response);
+          if (response.status === 200 && response.data.matchStatus === 'true') {
+            setMatchStatus('Matched');
+            clearInterval(interval);
+          }
+        })
+        .catch((error) => {
+          console.log(error);
```
kmh287, line 44: What does the user see if this happens? Should we show an error?

frontend/src/components/Form.js
```diff
@@ -42,10 +43,14 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+// Add onSubmit to props validation
+Form.propTypes = {
+  onSubmit: PropTypes.func,
+};
+
+
```
kmh287, line 51: tiny nit: remove extra line here

frontend/src/App.js
dhruvi618: Currently, added an alert to the screen along with a TO-DO (linked to an issue) to potentially add a page for that

frontend/src/components/Form.js
dhruvi618: Fixed, thanks

frontend/src/App.js
```diff
@@ -26,6 +27,26 @@ const useStyles = makeStyles((theme) => ({
  */
 export default function App() {
   const classes = useStyles();
+  const matchDataRefreshRate = 5000;
```
jzwang8, line 30: Our site is kinda slow/laggy after you submit now maybe because of refreshing too many times? We should investigate and maybe refresh less frequently

dhruvi618: Yes, I am going to further investigate on how it reacts once the multiple page views are added and adjust accordingly. Thank you

frontend/src/App.js
```diff
@@ -26,6 +27,26 @@ const useStyles = makeStyles((theme) => ({
...
+  const [matchStatus, setMatchStatus] = React.useState('Unmatched');
```
kmh287, line 31: Can "unmatched" and "matched" be pulled out to constants?

frontend/src/App.test.js
```diff
@@ -11,12 +12,28 @@ beforeAll(() => {
...
+    const servletEndpoint = '/api/v1/search-match';
+    const mockMatchStatus = {matchStatus: 'true'};
+    const mockData = {data: mockMatchStatus};
+
+    axios.get.mockImplementation(() => Promise.resolve(mockData));
+    const responseData = axios.get(servletEndpoint);
+
+    responseData.then((response) => {
+      expect(response.data).toEqual(mockData);
+    }).catch((error) => {});
```
kmh287, line 37: IIRC one has to do something special to test promises in javascript. It's possible that this test will run to completion without ever running the `then` block. You can test this out by adding a false assertion into the body of the `then` block and seeing if the test still passes.

frontend/src/App.test.js
```diff
@@ -11,12 +12,28 @@ beforeAll(() => {
...
     expect(tree).toMatchSnapshot();
   });
+
+  it('should initiate GET request using axios() to servlet', () => {
+    const servletEndpoint = '/api/v1/search-match';
+    const mockMatchStatus = {matchStatus: 'true'};
+    const mockData = {data: mockMatchStatus};
+
+    axios.get.mockImplementation(() => Promise.resolve(mockData));
+    const responseData = axios.get(servletEndpoint);
```
kmh287, line 33: This test should interact with the App in some way rather than mocking the network response and then sending the request from the test. Otherwise this test is only testing its own setup.

frontend/src/App.js
TcgVanguardTroll, line 31: Would you mind explaining this line to me ?

dhruvi618: Yes, pulled them out to constants

dhruvi618: Yes, this line is using the state hook in React. Essentially, it is declaring a new state variable for the matchStatus and a setter function for it. It is also setting the default match status to 'Unmatched' since the user is unmatched when first loading the application.

frontend/src/App.test.js
dhruvi618: Oh good point, changed it to interact with App instead of testing setup.

frontend/src/App.test.js
dhruvi618: I added the false assertion and saw that the test did run to completion without running the then block. I changed the test to use async await instead and also found that another way of doing this would involved adding done() to each of the then and catch blocks.

frontend/src/App.test.js
```diff
@@ -25,12 +26,24 @@ beforeAll(() => {
...
     expect(tree).toMatchSnapshot();
   });
+
+  it('should initiate GET request using axios() to servlet', async () => {
+    const mockMatchStatus = {matchStatus: 'true'};
+    const mockData = {data: mockMatchStatus};
+    axios.get.mockResolvedValue(mockData);
+
+    const response = await fetchMatch();
+    return expect(response).toEqual(mockMatchStatus);
```
kmh287, line 47: Does the test need to return the result of the assertion? The pattern I've seen with tests that assert on promises is to return the promise at the end of the test. Like

```
 it('should initiate GET request using axios() to servlet', () => {
    const mockMatchStatus = {matchStatus: 'true'};
    const mockData = {data: mockMatchStatus};
    axios.get.mockResolvedValue(mockData);

    return fetchMatch().then((response) => {
        expect(response).toEqual(mockMatchStatus);
    })
});
```

would that work here or does that not work with this framework?

frontend/src/App.test.js
kmh287: forgot to push?

frontend/src/App.test.js
dhruvi618: Yes, both of these would work with the framework since both await and then resolve the promise.

PR #80, opened 2020-07-28
Add jest-html tool to preview snapshot files in browser

dhruvi618:
Files changed: 
- All snapshot files have been changed to add HTML preview reference for snapshot tests
- yarn.lock - Update existing dependencies and plugins and add dependencies for jest-html
- package.json - Add jest-html dependency and yarn script to add functionality to run in browser

Example of jest-html snapshot browser preview: https://screenshot.googleplex.com/Zz4iE9jdqTq

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -30,7 +30,8 @@
     "local": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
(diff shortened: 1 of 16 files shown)


dhruvi618:
> Any idea why the files need a distinct "HTML preview" beyond what's already in the snapshot?

The documentation said that it "diffs" on the raw snapshots that we had before and the HTML previews

PR #81, opened 2020-07-29
Add frontend and backend functionality to remove participant from matching queue

dhruvi618:
Files changed:

Backend:
- RemoveParticipantServlet.java: Add servlet with doPost method which instantiates a RemoveParticipantHelper to remove a participant from the queue
- RemoveParticipantHelper.java: Implement functionality of doPost method to remove participant from datastore on request
- RemoveParticipantHelperTest.java: Add tests to ensure correct remove participant functionality
- ParticipantDatastore.java: Modify class to not be final so it can be mocked in tests
- UsernameService.java: Add class that encapsulates the getUsername() functionality as a wrapper for tests and so it can easily be used by other servlets.
- UsernameServiceTest.java: Add tests to ensure correct get username functionality

Frontend:
- App.js: Add ExitQueuePage redirect method to change page view when user successfully exits queue and add exit-queue option in switch statement
- ExitQueuePage.js: Add ExitQueuePage component with content to serve as confirmation that the user has exited the queue and can leave the page
- ExitQueuePage.test.js: Add snapshot UI tests to ensure ExitQueuePage component renders correctly
- LoadingPage.js: Update LoadingPage logic to add functionality to button click and send POST request to remove-participant-servlet
- LoadingPage.test.js: Update LoadingPage tests to add test on newly added POST request

Snapshot files changed:
- LoadingPage.test.js.snap: Update snapshot of LoadingPage component to account for onClick functionality for 'Exit Queue' button added.
- ExitQueuePage.test.js.snap: Create reference snapshot of ExitQueuePage tests


backend/src/main/java/com/google/sps/servlets/RemoveParticipantServlet.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that removes a participant from the queue */
+@WebServlet("/api/v1/remove-participant")
+public class RemoveParticipantServlet extends HttpServlet {
+
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final ParticipantDatastore participantDatastore = new ParticipantDatastore(datastore);
+  private final UsernameService usernameService = new UsernameService();
+  private final RemoveParticipantHelper helper =
```
jzwang8, line 35: Remove?

backend/src/main/java/com/google/sps/RemoveParticipantHelper.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Helper class to execute doPost for RemoveParticipantServlet */
+public final class RemoveParticipantHelper {
+
+  private final ParticipantDatastore participantDatastore;
+  private final UsernameService usernameService;
+  private static final String RESPONSE = "Received remove request.";
```
jzwang8, line 27: Why not have this hard coded into line 49?

backend/src/main/java/com/google/sps/UsernameService.java
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+
+/** Class that returns a username */
+public class UsernameService {
```
jzwang8, line 21: Super helpful/useful class

backend/src/main/java/com/google/sps/servlets/RemoveParticipantServlet.java
```diff
@@ -0,0 +1,42 @@
...
+      new RemoveParticipantHelper(participantDatastore, usernameService);
```
jzwang8, line 36: I put this in this class's doPost method with request/response parameters too, and had the helper doPost method take in request, response, datastores, and usernameService. Whatever we decide, we should keep it consistent across the servlets. Thoughts?

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+
+@RunWith(JUnit4.class)
+public final class RemoveParticipantHelperTest {
+
+  @Mock UsernameService usernameService = mock(UsernameService.class);
+  @Mock HttpServletRequest request = mock(HttpServletRequest.class);
+  @Mock HttpServletResponse response = mock(HttpServletResponse.class);
+  @Mock ParticipantDatastore participantDatastore = mock(ParticipantDatastore.class);
```
jzwang8, line 40: Another consistency comment. I currently have the mock(....class) part in a setUp method since I also use datastore in my tests.

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test
+  public void testDoPost_shouldRemoveParticipant() throws IOException {
+    when(usernameService.getUsername()).thenReturn(USER);
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    helper.doPost(request, response);
+
+    writer.flush();
```
jzwang8, line 57: Is it necessary to flush? I didn't so am interested to hear your rationale in including

backend/src/main/java/com/google/sps/UsernameService.java
dhruvi618: Thanks!

backend/src/main/java/com/google/sps/RemoveParticipantHelper.java
dhruvi618: Since I took it out as a constant in the RemoveParticipantHelperTest, I decided to do the same thing in this class as well.

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
dhruvi618: Since there is only one functionality being tested in this test class, I don't think there is a need for a setUp method here and it can instead be done in-line within the declaration.

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
dhruvi618: In the current state of the class, there is not a need for this flush(), however, I think in the case where this test class expands and the same writer is used in numerous tests, this flush functionality would help in clearing out the writer before it is asserted to contain a certain value.

backend/src/main/java/com/google/sps/servlets/RemoveParticipantServlet.java
dhruvi618: The rationale for taking this out of the doPost method was so that the servlet could instantiate this once and reuse it. I think this would help in future for scalability reasons, what do you think?

backend/src/main/java/com/google/sps/RemoveParticipantHelper.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public RemoveParticipantHelper(
+      ParticipantDatastore participantDatastore, UsernameService usernameService) {
+    this.participantDatastore = participantDatastore;
+    this.usernameService = usernameService;
+  }
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
```
kmh287, line 36: tiny nit: remove extra newline here

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
kmh287: Either way works. My only suggestion here would be to remove the @Mock annotation and make these private final fields of the test. The @Mock annotation is meant to work with a call to Mockito.initMocks(testClassInstance) which is usually placed in setUp and is an alternative to manually calling mock as you have here. Without a call to initMocks the @Mock annotation does nothing.

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
```diff
@@ -37,7 +37,7 @@
 import javax.annotation.Nullable;
 
 /** Separates datastore method calls involving Participant type from caller */
-public final class ParticipantDatastore {
+public class ParticipantDatastore {
```
TcgVanguardTroll, line 40: Quick question, this class is no longer final because ?

backend/src/test/java/com/google/sps/UsernameServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  public void testGetUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+
+    String email = userService.getCurrentUser().getEmail();
+    String username = email.split("@")[0];
+
+    assertEquals(email, EMAIL);
+    assertEquals(username, USER);
+  }
```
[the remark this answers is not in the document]
TcgVanguardTroll: Do you think making username service take as a parameter user service would be a good option ?

backend/src/main/java/com/google/sps/datastore/ParticipantDatastore.java
dhruvi618: Great question, this class is no longer final because it needed to be mocked in the tests and final classes cannot be mocked.

backend/src/test/java/com/google/sps/UsernameServiceTest.java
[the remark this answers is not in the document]
kmh287: Definitely 👍

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
dhruvi618: Removed the @Mock annotations, thank you for the suggestion!

backend/src/main/java/com/google/sps/servlets/RemoveParticipantServlet.java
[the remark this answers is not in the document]
dhruvi618: Great suggestion, thank you very much!

backend/src/test/java/com/google/sps/UsernameServiceTest.java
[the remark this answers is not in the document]
dhruvi618: Yes, I think that is a great idea! I added that in, thank you for the suggestion!

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class RemoveParticipantHelperTest {
+
+  UsernameService usernameService = mock(UsernameService.class);
+  HttpServletRequest request = mock(HttpServletRequest.class);
+  HttpServletResponse response = mock(HttpServletResponse.class);
+  ParticipantDatastore participantDatastore = mock(ParticipantDatastore.class);
```
kmh287, line 39: now these can be private final (they couldn't previously with the @Mock annotation)

backend/src/test/java/com/google/sps/RemoveParticipantHelperTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private RemoveParticipantHelper helper =
+      new RemoveParticipantHelper(participantDatastore, usernameService);
+  private StringWriter stringWriter = new StringWriter();
+
+  private static final String USER = "user";
+  private static final String EXPECTED_RESPONSE = "Received remove request.";
+
```
kmh287, line 47: Could we also test the case where the user's username isn't found from the Username service?

dhruvi618: Added, thanks!

PR #82, opened 2020-07-29
Move match preference string to enum conversion to MatchPreference

jzwang8:
Files changed:
- MatchPreference.java: add forStringValue for conversion from string to enum
- AddParticipantServlet.java: remove conversion method

### DIFF
--- a/backend/src/main/java/com/google/sps/data/MatchPreference.java
+++ b/backend/src/main/java/com/google/sps/data/MatchPreference.java
@@ -41,4 +41,17 @@ public enum MatchPreference {
     }
     throw new IllegalStateException("Unknown enum value.");
   }
+
+  public static MatchPreference forStringValue(String value) {
+    switch (value) {
+      case "different":
+        return MatchPreference.DIFFERENT;
+      case "any":
+        return MatchPreference.ANY;
+      case "similar":
+        return MatchPreference.SIMILAR;
+      default:
+        return null;
+    }
+  }
 }
--- a/backend/src/main/java/com/google/sps/data/MatchPreference.java
+++ b/backend/src/main/java/com/google/sps/data/MatchPreference.java
@@ -39,11 +39,11 @@ public enum MatchPreference {
         return preference;
       }
     }
-    throw new IllegalStateException("Unknown enum value.");
+    throw new IllegalStateException("Unknown enum int value.");
   }
 
   public static MatchPreference forStringValue(String value) {
-    switch (value) {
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -41,4 +41,17 @@ public static MatchPreference forIntValue(int value) {
...
+    switch (value) {
+      case "different":
+        return MatchPreference.DIFFERENT;
+      case "any":
+        return MatchPreference.ANY;
+      case "similar":
+        return MatchPreference.SIMILAR;
+      default:
+        return null;
+    }
```
kmh287, line 55: There's a good opportunity for a test here. How should this method handle "ANY" or "SiMiLaR"?

backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -41,4 +41,17 @@ public static MatchPreference forIntValue(int value) {
...
+  public static MatchPreference forStringValue(String value) {
+    switch (value) {
+      case "different":
+        return MatchPreference.DIFFERENT;
+      case "any":
+        return MatchPreference.ANY;
+      case "similar":
+        return MatchPreference.SIMILAR;
+      default:
+        return null;
```
tcwanggoogle, line 54: Should we throw exception here like the forIntValue()? Otherwise, anywhere you want to use MatchPreference, you have to check for null.

backend/src/main/java/com/google/sps/data/MatchPreference.java
jzwang8: I added toLowerCase() to value, but also value should always be one of those three strings since that's what the frontend passes, so if it's none of the above, then an error should be thrown

kmh287: Looks good. There's still an opportunity here for a (short) unit test.

jzwang8: Added tests

kmh287:
Nice job on the tests

jzwang8:
> LGTM! One general comment, since we have added a UsernameService class that now implements the getUsername functionality, WDYT about taking the functionality out of this class? I think we can wait and do it in a follow up PR so it's consistent in all servlets if that works better.

Yep, I created this PR before we refactored servlets/tests. Other servlet PRs have the UsernameService class since that's not in master yet and they all depend on that.

PR #83, opened 2020-07-29
Add backend support for loading user preferences

jzwang8:
File added:
- LoadUserPreferencesServlet: return JSON object of current user when called

### DIFF
--- a/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
+++ b/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
@@ -14,6 +14,7 @@
 
 package com.google.sps;
 
+import com.google.sps.data.MatchPreference;
 import com.google.sps.data.User;
 import com.google.sps.datastore.UserDatastore;
 import java.io.IOException;
@@ -27,7 +28,6 @@ public class LoadUserPreferencesHelper {
 
   // JSON key constants
   private static final String JSON_EXISTING = "existing";
-  private static final String JSON_USERNAME = "username";
   private static final String JSON_DURATION = "duration";
   private static final String JSON_ROLE = "role";
   private static final String JSON_PRODUCT_AREA = "productArea";
@@ -76,7 +76,7 @@ public class LoadUserPreferencesHelper {
     userJson.put(JSON_ROLE, user.getRole());
     userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
     userJson.put(JSON_INTERESTS, new JSONArray(user.getInterests()));
-    userJson.put(JSON_MATCH_PREFERENCE, user.getMatchPreference().getValue());
+    userJson.put(JSON_MATCH_PREFERENCE, MatchPreference.getStringValue(user.getMatchPreference()));
 
     // Send the JSON back as the response
     response.setContentType("application/json");
--- a/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
+++ b/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
@@ -34,32 +34,18 @@ public class LoadUserPreferencesHelper {
   private static final String JSON_INTERESTS = "interests";
   private static final String JSON_MATCH_PREFERENCE = "matchPreference";
 
-  // HttpServlet request and response
-  private final HttpServletRequest request;
-  private final HttpServletResponse response;
-  // User Datastore
   private final UserDatastore userDatastore;
-
   private final UsernameService usernameService;
 
   /** Constructor */
-  public LoadUserPreferencesHelper(
-      HttpServletRequest request,
-      HttpServletResponse response,
-      UserDatastore userDatastore,
-      UsernameService usernameService) {
-    this.request = request;
-    this.response = response;
+  public LoadUserPreferencesHelper(UserDatastore userDatastore, UsernameService usernameService) {
     this.userDatastore = userDatastore;
     this.usernameService = usernameService;
   }
 
   /** Returns a JSON object of the user's preferences */
-  public void doGet() throws IOException {
-
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     System.out.println("Request received");
-
-    JSONObject userJson = new JSONObject();
 
     // Find participant's match, if exists and not returned yet
     String username = usernameService.getUsername();
@@ -67,6 +53,8 @@ public class LoadUserPreferencesHelper {
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not retrieve email");
       return;
     }
+
+    JSONObject userJson = new JSONObject();
 
     User user = userDatastore.getUserFromUsername(username);
     if (user == null) {
--- a/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
+++ b/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
@@ -61,7 +61,7 @@ public class LoadUserPreferencesHelper {
       response.setStatus(
           HttpServletResponse.SC_OK,
           "No saved preferences for user with username " + username + ".");
-      userJson.put(JSON_EXISTING, "false");
+      userJson.put(JSON_EXISTING, false);
       response.setContentType("application/json");
       response.getWriter().println(userJson.toString());
       return;
@@ -71,7 +71,7 @@ public class LoadUserPreferencesHelper {
         HttpServletResponse.SC_OK,
         "Loading saved preferences for user with username " + username + ".");
 
-    userJson.put(JSON_EXISTING, "true");
+    userJson.put(JSON_EXISTING, true);
     userJson.put(JSON_DURATION, user.getDuration());
     userJson.put(JSON_ROLE, user.getRole());
     userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
--- a/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
+++ b/backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
@@ -1,0 +1,97 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import com.google.sps.data.User;
+import com.google.sps.datastore.UserDatastore;
+import java.io.IOException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+/** Helper for LoadUserPreferencesServlet */
+public class LoadUserPreferencesHelper {
+
+  // JSON key constants
+  private static final String JSON_EXISTING = "existing";
+  private static final String JSON_USERNAME = "username";
+  private static final String JSON_DURATION = "duration";
+  private static final String JSON_ROLE = "role";
+  private static final String JSON_PRODUCT_AREA = "productArea";
+  private static final String JSON_INTERESTS = "interests";
+  private static final String JSON_MATCH_PREFERENCE = "matchPreference";
+
+  // HttpServlet request and response
+  private final HttpServletRequest request;
+  private final HttpServletResponse response;
+  // User Datastore
+  private final UserDatastore userDatastore;
+
+  private final UsernameService usernameService;
+
+  /** Constructor */
+  public LoadUserPreferencesHelper(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      UserDatastore userDatastore,
+      UsernameService usernameService) {
+    this.request = request;
+    this.response = response;
+    this.userDatastore = userDatastore;
+    this.usernameService = usernameService;
+  }
+
+  /** Returns a JSON object of the user's preferences */
+  public void doGet() throws IOException {
+
+    System.out.println("Request received");
+
+    JSONObject userJson = new JSONObject();
+
+    // Find participant's match, if exists and not returned yet
+    String username = usernameService.getUsername();
+    if (username == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not retrieve email");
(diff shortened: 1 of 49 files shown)


backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.annotation.Nullable;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.simple.JSONObject;
+
+/** Servlet that returns a JSON object of the user's preferences */
+@WebServlet("/api/v1/load-user")
+public class LoadUserPreferencesServlet extends HttpServlet {
```
tcwanggoogle, line 33: nit: (and more like a question for Java servlet) It looks like the static strings here and the member function getUsername() are similar to AddParticipant servlet. Can we inherit that servlet and only override doGet() function? (And much of the doGet() function can be of its own function in AddParticipant, so this class can be just calling that function plus the return of userJson)

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    System.out.println("Request received");
+
+    // Find participant's match, if exists and not returned yet
+    User user = userDatastore.getUserFromUsername(getUsername());
+    if (user == null) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
```
kmh287, line 58: Is this case a bad request though? This seems like the standard flow for a user who has never stored preferences before

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,84 @@
...
+  // Get user username
+  private final UserService userService = UserServiceFactory.getUserService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    System.out.println("Request received");
+
+    // Find participant's match, if exists and not returned yet
+    User user = userDatastore.getUserFromUsername(getUsername());
```
kmh287, line 55: Do you need to check `getUsername()` for null?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
dhruvi618: +1. Can we send back a non-error response for this?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,84 @@
...
+    if (user == null) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
+          "User with username " + getUsername() + "does not exist in datastore.");
+      return;
+    }
+    JSONObject userJson = new JSONObject();
+    userJson.put(JSON_USERNAME, user.getUsername());
```
dhruvi618, line 63: Since we don't currently have a username field that a user inputs on the frontend, I don't know if this definitely needs to be sent back

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,84 @@
...
+    userJson.put(JSON_DURATION, user.getDuration());
+    userJson.put(JSON_ROLE, user.getRole());
+    userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
+    userJson.put(JSON_INTERESTS, UserDatastore.convertListToString(user.getInterests()));
```
dhruvi618, line 67: Is it possible to send these back as the array that was sent over on form submission?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
jzwang8: Yep, now returning JSON object with field "existing" = "true" if user exists, "false" if not

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
jzwang8: WDYT about adding a new class called GetUsername.java or something that has the one getUsername() method, which all the servlets can call? so not tying that method directly to only AddParticipantServlet, since that seems a bit arbitrary.

kmh287: Delegation to a helper class could work. Another option would be a static utility class that takes an email String and returns the username

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,90 @@
...
+      response.getWriter().println(userJson.toString());
+      return;
+    } else {
+      userJson.put(JSON_EXISTING, "true");
+      userJson.put(JSON_DURATION, user.getDuration());
+      userJson.put(JSON_ROLE, user.getRole());
+      userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
+      userJson.put(JSON_INTERESTS, new JSONArray(user.getInterests()));
+      userJson.put(JSON_MATCH_PREFERENCE, user.getMatchPreference().getValue());
+    }
```
kmh287, line 78: since the if returns, block can move out of the else

jzwang8:
> Please add tests for the new servlet.

Added tests

backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
```diff
@@ -0,0 +1,97 @@
...
+    response.setStatus(
+        HttpServletResponse.SC_OK,
+        "Loading saved preferences for user with username " + username + ".");
+
+    userJson.put(JSON_EXISTING, "true");
+    userJson.put(JSON_DURATION, user.getDuration());
+    userJson.put(JSON_ROLE, user.getRole());
+    userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
+    userJson.put(JSON_INTERESTS, new JSONArray(user.getInterests()));
+    userJson.put(JSON_MATCH_PREFERENCE, user.getMatchPreference().getValue());
```
dhruvi618, line 91: Wondering if these methods will return an empty string or null if a specific value is not stored in the datastore?

backend/src/test/java/com/google/sps/LoadUserPreferencesHelperTest.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final int DURATION_DEFAULT = 30;
+  private static final String ROLE_DEFAULT = "Software engineer";
+  private static final String PRODUCT_AREA_DEFAULT = "Ads";
+  private static final List<String> INTERESTS_DEFAULT = Arrays.asList("Books");
+  private static final MatchPreference MATCH_PREFERENCE_DEFAULT = MatchPreference.ANY;
+
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
+  @Mock private UserDatastore userDatastore;
+  @Mock private UsernameService usernameService;
```
kmh287, line 40: nit: the @Mock annotation is meant to be used with Mockito.initMocks(myTestClass). One can either initialize mocks manually by calling the mock method or automatically by using this annotation and calling initMocks on the test class. If initMocks sin't used the annotation does nothing.

I'm partial to manually calling mock and omitting the annotation because it ends up being cleaner in kotlin but you could go either way.

backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
jzwang8: Empty string!/list with 0 items

backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
```diff
@@ -0,0 +1,97 @@
...
+    this.userDatastore = userDatastore;
+    this.usernameService = usernameService;
+  }
+
+  /** Returns a JSON object of the user's preferences */
+  public void doGet() throws IOException {
+
+    System.out.println("Request received");
+
+    JSONObject userJson = new JSONObject();
```
kmh287, line 62: Can this be declared closer to where it's used (after the username check)?

backend/src/test/java/com/google/sps/LoadUserPreferencesHelperTest.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void noUserWithUsernameInDatastore() throws IOException {
+    when(response.getWriter()).thenReturn(getWriter());
```
kmh287, line 63: This looks like common setup for all the tests. Consider moving it into the setUp method.

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,44 @@
...
+  // Get DatastoreService and instantiate User Datastore
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final UserDatastore userDatastore = new UserDatastore(datastore);
+
+  private final UsernameService usernameService = new UsernameService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    LoadUserPreferencesHelper loadUserPreferencesHelper =
+        new LoadUserPreferencesHelper(request, response, userDatastore, usernameService);
```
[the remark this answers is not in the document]
jzwang8: Good point, I'll change this in the other servlets/tests too

backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+/** Helper for LoadUserPreferencesServlet */
+public class LoadUserPreferencesHelper {
+
+  // JSON key constants
+  private static final String JSON_EXISTING = "existing";
+  private static final String JSON_USERNAME = "username";
```
dhruvi618, line 30: remove?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,44 @@
...
+  // Get DatastoreService and instantiate User Datastore
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final UserDatastore userDatastore = new UserDatastore(datastore);
+
+  private final UsernameService usernameService = new UsernameService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    LoadUserPreferencesHelper loadUserPreferencesHelper =
+        new LoadUserPreferencesHelper(userDatastore, usernameService);
```
kmh287, line 41: now this can be a field on the servlet

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
```diff
@@ -0,0 +1,44 @@
...
+
+/** Servlet that returns a JSON object of the user's preferences */
+@WebServlet("/api/v1/load-user")
+public class LoadUserPreferencesServlet extends HttpServlet {
+
+  // Get DatastoreService and instantiate User Datastore
+  private final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private final UserDatastore userDatastore = new UserDatastore(datastore);
+
+  private final UsernameService usernameService = new UsernameService();
```
kmh287, line 36: once the helper becomes a field on the servlet, these don't need to be fields anymore

backend/src/test/java/com/google/sps/LoadUserPreferencesHelperTest.java
```diff
@@ -0,0 +1,101 @@
...
+    verify(response)
+        .setStatus(
+            HttpServletResponse.SC_OK,
+            "Loading saved preferences for user with username " + USERNAME_PERSON_A + ".");
+  }
+
+  /** Mock HttpServletResponse method */
+  private PrintWriter getWriter() {
+    return new PrintWriter(System.out);
+  }
```
kmh287, line 100: now that this is used once, WDYT about inlining it?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
tcwanggoogle: Sure help class wfm. My comment was more like seeing a lot repeated code, including the private variables, etc. and trying to think of ways to optimize. If the helper is not used by other servlets (or is it?), it's the same with or without helper?

backend/src/main/java/com/google/sps/LoadUserPreferencesHelper.java
```diff
@@ -0,0 +1,85 @@
...
+    response.setStatus(
+        HttpServletResponse.SC_OK,
+        "Loading saved preferences for user with username " + username + ".");
+
+    userJson.put(JSON_EXISTING, true);
+    userJson.put(JSON_DURATION, user.getDuration());
+    userJson.put(JSON_ROLE, user.getRole());
+    userJson.put(JSON_PRODUCT_AREA, user.getProductArea());
+    userJson.put(JSON_INTERESTS, new JSONArray(user.getInterests()));
+    userJson.put(JSON_MATCH_PREFERENCE, user.getMatchPreference().getValue());
```
dhruvi618, line 79: Is it possible to return the string value of the matchPreference here?

backend/src/main/java/com/google/sps/servlets/LoadUserPreferencesServlet.java
jzwang8: I'm not sure I understand what you mean. LoadUserPreferencesHelper takes in userDatastore and usernameService as parameters, so wouldn't this also need to be declared and initialized here but before the declaration of the helper?

kmh287: They can be created inline like this:

```
  private final LoadUserPreferencesHelper loadUserPreferencesHelper =
      new LoadUserPreferencesHelper(
          new UserDatastore(DatastoreServiceFactory.getDatastoreService()),
          new UsernameService());
```

PR #84, opened 2020-07-29
Add servlet for cron job to remove expired participants

jzwang8:
Files changed:
- ParticipantDatastore: Add method to return list of unmatched participants
- RemoveExpiredParticipantsServlet: Remove expired participants from datastore when called
- cron.xml: add cron job to call doGet of RemoveExpiredParticipantsServlet every 1 hour


backend/src/main/java/com/google/sps/servlets/RemoveExpiredParticipantsServlet.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    System.out.println("Remove expired participants");
+
+    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
+
+    // Check if match exists and not returned yet
+    for (Participant participant : unmatchedParticipants) {
+      if (participant.getMatchStatus() == MatchStatus.UNMATCHED && isExpired(participant)) {
+        participantDatastore.removeParticipant(participant.getUsername());
```
kmh287, line 50: Does the DataStoreService support a bulk delete operation? If so it might be good to expose that in ParticipantDatastore and then use it here to remove the participants in one transaction.

jzwang8: I think you may be able to delete a batch of entries at once, which would be helpful if we could query datastore for unmatched AND endTimeAvailable. Or else we'd have to loop through the list of unmatchedParticipants and create another list of the ones who are unmatched and expired. I couldn't find detailed documentation about that operation, though. And I was hesitant to try out querying by endTimeAvailable since I had so much trouble doing that last time--I wanted to see if the cron job works before optimizing the query.

PR #85, opened 2020-07-30
Add servlet for cron job to remove expired participants

jzwang8:
Files changed:
- ParticipantDatastore: Add method to return list of unmatched participants
- RemoveExpiredParticipantsServlet: Remove expired participants from datastore when called
c- ron.xml: add cron job to call doGet of RemoveExpiredParticipantsServlet every 1 hour

### DIFF
--- a/backend/src/main/java/com/google/sps/servlets/RemoveExpiredParticipantsServlet.java
+++ b/backend/src/main/java/com/google/sps/servlets/RemoveExpiredParticipantsServlet.java
@@ -1,0 +1,70 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/google/sps/servlets/RemoveExpiredParticipantsServlet.java
```diff
@@ -0,0 +1,70 @@
...
+    List<Participant> unmatchedParticipants = participantDatastore.getUnmatchedParticipants();
+
+    // Check if match exists and not returned yet
+    for (Participant participant : unmatchedParticipants) {
+      if (isExpired(participant)) {
+        participantDatastore.removeParticipant(participant.getUsername());
+        return;
+      }
+    }
+    response.setStatus(HttpServletResponse.SC_OK);
```
kmh287, line 53: Continuing from the other PR:

> I think you may be able to delete a batch of entries at once, which would be helpful if we could query datastore for unmatched AND endTimeAvailable. Or else we'd have to loop through the list of unmatchedParticipants and create another list of the ones who are unmatched and expired. I couldn't find detailed documentation about that operation, though. And I was hesitant to try out querying by endTimeAvailable since I had so much trouble doing that last time--I wanted to see if the cron job works before optimizing the query.

I agree that the query is sufficient for now. But at this step, it would still make sense to delete all of the expired participants in one delete operation instead of one transaction per participant. 
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/DatastoreService.html#delete-java.lang.Iterable-

PR #87, opened 2020-07-31
Add tests for AddParticipantServlet and SearchMatchServlet and add helper classes

jzwang8:
Files changed/added:
- AddParticipantServlet: moved all functionality to AddParticipantHelper 
- AddParticipantHelper: implement add participant logic
- AddParticipantHelperTest: add tests for helper class
- SearchMatchServlet: moved all functionality to SearchMatchServlet
- SearchMatchHelper: implement search match logic
- SearchMatchHelperTest: add tests for helper class
- MatchDatastore/ParticipantDatastore/UserDatastore: make not final for mocking
- MatchPreference: add forStringValue method


backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -41,4 +41,18 @@ public static MatchPreference forIntValue(int value) {
...
+      case "different":
+        return MatchPreference.DIFFERENT;
+      case "any":
+        return MatchPreference.ANY;
+      case "similar":
+        return MatchPreference.SIMILAR;
+      default:
+        throw new IllegalStateException("Unknown enum value.");
+    }
+  }
```
kmh287, line 57: This was submitted already in another PR right?

backend/src/main/java/com/google/sps/servlets/AddParticipantServlet.java
```diff
@@ -16,171 +16,38 @@
...
-      default:
-        return null;
-    }
+    AddParticipantHelper addParticipantHelper =
+        new AddParticipantHelper(
+            Clock.systemUTC(),
+            matchDatastore,
+            participantDatastore,
+            userDatastore,
+            usernameService);
```
kmh287, line 50: This can be reused if it becomes a field on this class.

backend/src/main/java/com/google/sps/servlets/SearchMatchServlet.java
```diff
@@ -16,183 +16,31 @@
...
-  /** Send email notification when participant is expired */
-  private void sendExpiredEmail() {
-    try {
-      emailNotifier = createOrGetEmailNotifier();
-      emailNotifier.sendExpiredEmail(getUsername(), userService.getCurrentUser().getEmail());
-    } catch (GeneralSecurityException | MessagingException | IOException e) {
-      e.printStackTrace();
-    }
+    SearchMatchHelper searchMatchHelper =
+        new SearchMatchHelper(matchDatastore, participantDatastore, usernameService);
```
kmh287, line 43: make this a field?

backend/src/test/java/com/google/sps/AddParticipantHelperTest.java
```diff
@@ -0,0 +1,348 @@
...
+
+  /** Mock HttpServletRequest method */
+  private BufferedReader getReader(JSONObject jsonObject) {
+    return new BufferedReader(new StringReader(jsonObject.toString()));
+  }
+
+  /** Mock HttpServletResponse method */
+  private PrintWriter getWriter() {
+    return new PrintWriter(System.out);
+  }
```
kmh287, line 347: WDYT about inlining this?

backend/src/test/java/com/google/sps/SearchMatchHelperTest.java
```diff
@@ -0,0 +1,208 @@
...
+        new SearchMatchHelper(matchDatastore, participantDatastore, usernameService);
+    searchMatchHelper.doGet(request, response);
+
+    verify(response).setStatus(HttpServletResponse.SC_OK, "Participant has a match!");
+  }
+
+  /** Mock HttpServletResponse method */
+  private PrintWriter getWriter() {
+    return new PrintWriter(System.out);
+  }
```
kmh287, line 207: inline?

backend/src/test/java/com/google/sps/SearchMatchHelperTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void noParticipantInDatastore() throws IOException {
+    searchMatchHelper =
+        new SearchMatchHelper(matchDatastore, participantDatastore, usernameService);
```
kmh287, line 83: Since this happens in all the tests, can it move into the setUp method?

backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -41,4 +41,18 @@ public static MatchPreference forIntValue(int value) {
     }
     throw new IllegalStateException("Unknown enum value.");
   }
+
+  /** Parse match preference string to MatchPreference */
+  public static MatchPreference forStringValue(String value) {
+    switch (value.toLowerCase()) {
```
TcgVanguardTroll, line 47: NIt : do you think the lower case value should be made into a variable so that the function doesn't have to be called each time

backend/src/test/java/com/google/sps/SearchMatchHelperTest.java
jzwang8: Even though it happens in all the tests, 4/5 tests don't actually use the mock datastores since testing the different paths through the servlet requires actually putting entities into the datastore.

backend/src/main/java/com/google/sps/data/MatchPreference.java
jzwang8: Yes but I needed it for this PR so I just added it so it would build.

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,210 @@
...
+    if (obj == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Could not read request body");
+      return;
+    }
+    JSONObject formDetails = obj.getJSONObject(REQUEST_FORM_DETAILS);
+
+    // Get new Participant from input parameters
+    Participant newParticipant = getParticipantFromInputs(response, formDetails);
+    if (newParticipant == null) {
+      return;
```
kmh287, line 86: We'll want to send a response back to the client here

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,210 @@
...
+    return new Participant(
+        username,
+        startTimeAvailable,
+        endTimeAvailable,
+        duration,
+        role,
+        productArea,
+        interests,
+        matchPreference,
+        /* matchId=*/ 0,
```
kmh287, line 178: tiny nit: space after the = so that it's `/* matchID= */`

backend/src/main/java/com/google/sps/data/MatchPreference.java
```diff
@@ -42,16 +42,18 @@ public static MatchPreference forIntValue(int value) {
...
+    switch (lowerCasePreference) {
       case "different":
         return MatchPreference.DIFFERENT;
       case "any":
         return MatchPreference.ANY;
       case "similar":
         return MatchPreference.SIMILAR;
       default:
-        throw new IllegalStateException("Unknown enum String value.");
+        throw new IllegalStateException("Unknown enum value.");
```
kmh287, line 56: while we're here, let's concatenate `value` to the exception reason. That way if we ever get an unexpected value we know what it is and it'll be easier to debug.

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,212 @@
...
+      System.out.println("match not found");
+    }
+
+    // Confirm received form input
+    response.setContentType("text/plain;charset=UTF-8");
+    response.getWriter().println("Received form input details and queried!");
+  }
+
+  /** Retrieve JSON body payload and convert to a JSONObject for parsing purposes */
+  private JSONObject retrieveRequestBody(HttpServletRequest request) throws IOException {
```
kmh287, line 125: does this method still throw IOException or would any exception be caught below?

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,212 @@
...
+    StringBuilder requestBuffer = new StringBuilder();
+    try {
+      BufferedReader reader = request.getReader();
+      String currentLine;
+      while ((currentLine = reader.readLine()) != null) {
+        requestBuffer.append(currentLine);
+      }
+    } catch (IOException e) {
+      return null;
```
kmh287, line 134: Please add a Nullable annotation and a short description in the javadocs for when this would return null.

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+    return new JSONObject(requestBuffer.toString());
+  }
+
+  /** Get a Participant from form inputs */
+  private Participant getParticipantFromInputs(HttpServletResponse response, JSONObject formDetails)
```
kmh287, line 140: Nullable + docs

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /** Retrieve user email address via Users API and parse for username */
+  public static String getUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
+
+  /** Get list of Strings from JSONArray */
+  public static List<String> getListFromJsonArray(JSONArray array) {
```
kmh287, line 193: private?

backend/src/main/java/com/google/sps/AddParticipantHelper.java
```diff
@@ -0,0 +1,212 @@
...
+        MatchStatus.UNMATCHED,
+        timestamp);
+  }
+
+  /** Retrieve user email address via Users API and parse for username */
+  public static String getUsername() {
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
+    return email != null ? email.split("@")[0] : null;
+  }
```
dhruvi618, line 190: Since the UsernameService class is in master now, WDYT about removing this and method and calling it via a UsernameService instead?

backend/src/main/java/com/google/sps/AddParticipantHelper.java
jzwang8: Actually, I don't need this method anymore because it's in UsernameService now

backend/src/main/java/com/google/sps/AddParticipantHelper.java
jzwang8: Yes, it already uses UsernameService and never calls this getUsername(). I just forgot to remove it!

